PR #11, opened 2014-11-09
prefer IANA MIME for .ico

jomo:
`image/x-icon` is not a valid MIME type according to the IANA database.
the correct MIME type for .ico files is `image/vnd.microsoft.icon`

### DIFF
--- a/db.json
+++ b/db.json
@@ -5309,7 +5309,9 @@
     "source": "iana"
   },
   "image/vnd.microsoft.icon": {
-    "source": "iana"
+    "source": "iana",
+    "compressible": true,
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
Interesting discussion here! Very cool!

Fishrock123:
I think this can be closed. Again, we should serve what actually works for clients in these (unfortunate) cases.

dougwilson:
Well, I didn't close this because it will (and needs to be) fixed. I just need to implement the mechanism that allows users to set their preferred order (though until such time, we would continue to prefer the non-IANA type for ico thanks to the history of this extension).

PR #18, opened 2015-02-04
Remove dependency on node-mime module

broofa:
I'm switching node-mime over to mime-db. See https://github.com/broofa/node-mime/pull/107 .  Thus removing the [circular] dependency that will cause between these two modules is a Good Thing.  See also @jonathanong's comment to that effect at https://github.com/jshttp/mime-db/blob/master/scripts/build.js#L22


dougwilson:
Neat. What's going to happen to `node-mime`? Is is going to stay the same, but simply source from here?

dougwilson:
> So, really, at this point the only thing node-mime will offer over something like mime-types is that the types.json file it generates is a little more compact, and it allows for project-specific customization via define().

Ok, sounds good to me :)

broofa:
> What's going to happen to node-mime?  Is is going to stay the same, but simply source from here?

Yeah, see referenced pull request.  node-mime gets an install script that builds `types.json`, an only-whats-needed version of mime-db's `db.json`. This conversion seemed prudent given that the impetus for all this was users' desire for node-mime to work with browserify (i.e. in the browser where download sizes are important)... and types.json comes in at 8KB gzip'ed .vs. db.json's 17KB.

dougwilson:
I plan to release this tonight so I'm not blocking you :)

dougwilson:
Thanks, @broofa , this has been published as 1.6.1. I also added you as a contributor: https://github.com/jshttp/mime-db/blob/v1.6.1/package.json#L8

jonathanong:
:+1

broofa:
Just published 1.3.2, which includes the aforementioned PR.

Thanks for the contributor nod!

PR #21, opened 2015-02-23
fix missing application/ld+json extension

dlongley:
This adds the `jsonld` extension to mime type "application/ld+json". See: http://www.iana.org/assignments/media-types/application/ld+json.

### DIFF
--- a/src/custom.json
+++ b/src/custom.json
@@ -59,6 +59,9 @@
   "application/json5": {
     "extensions": ["json5"]
   },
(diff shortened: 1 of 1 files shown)


dougwilson:
Thank you! Eventually we'll start pulling in the extensions from IANA directly, but this will tide us over :)

dlongley:
Sure! ... and automating pulling in extensions sounds great. :)

Issue #6, opened 2014-08-30
source.md from mime-types


jonathanong:
### Sources for custom types

This is a list of sources for any custom mime types.
When adding custom mime types, please link to where you found the mime type,
even if it's from an unofficial source.

- `text/coffeescript` - http://coffeescript.org/#scripts
- `text/x-handlebars-template` - https://handlebarsjs.com/#getting-started
- `text/x-sass` & `text/x-scss` - https://github.com/janlelis/rubybuntu-mime/blob/master/sass.xml
- `text.jsx` - http://facebook.github.io/react/docs/getting-started.html [[2]](https://github.com/facebook/react/blob/f230e0a03154e6f8a616e0da1fb3d97ffa1a6472/vendor/browser-transforms.js#L210)

[Sources for node.json types](https://github.com/broofa/node-mime/blob/master/types/node.types)

### Notes on weird types

- `font/opentype` - This type is technically invalid according to the spec. No valid types begin with `font/`. No-one uses the official type of `application/vnd.ms-opentype` as the community standardized `application/x-font-otf`. However, chrome logs nonsense warnings unless opentype fonts are served with `font/opentype`. [[1]](http://stackoverflow.com/questions/2871655/proper-mime-type-for-fonts)

dougwilson:
We don't really need endless issues; if there is something to address here, let me know :)

PR #23, opened 2015-03-27
Added one more BMP mime type

mikhail-guru:


jonathanong:
do you have a sourcE?

mikhailguru:
@jonathanong  Source was a libmime library, i.e. if you will take BMP file (at least which I have) and do `file this_file.bmp` command you get this source type. I've found such file on http://atheos.cx/screenshots/shot7.php3. Russian Wikipedia (http://ru.wikipedia.org/wiki/BMP) tells us that such mimetype exists, and `file` command also detects it as "image/x-ms-bmp".

dougwilson:
I agree that I have seem `image/x-ms-bmp` around in places.

jonathanong:
hmmm i swore we had a `sources.md` file or something. oh well

jonathanong:
oh. https://github.com/jshttp/mime-db/issues/6

dougwilson:
For now we should be adding a `sources` property to the object with an array of strings defining source(s).

PR #27, opened 2015-06-03
Added BDOC types

svenvarkel:
I added 2 new types: application/bdoc and application/x-bdoc that are used interchangeably for Digital Signature containers in the EU.

### DIFF
--- a/src/custom-suffix.json
+++ b/src/custom-suffix.json
@@ -2,9 +2,5 @@
   "+json": {
     "compressible": true,
     "notes": "+json means it is JSON, which is compressible."
-  },
-  "bdoc": {
-    "compressible": false,
-    "notes": "Digidoc (digital Signature container)."
   }
 }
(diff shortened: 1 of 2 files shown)


dougwilson:
Thank you! I'm not sure if I'm missing something, but I don't seem to find the MIME listed in the sources? One of the PDFs just won't load from my connection (getting connection reset), but the other doesn't seem to list it. Do you know what page it's on in which PDF?

svenvarkel:
Hi,

the link http://wpki.eu/wiki/upload/4/4f/BDoc-1.02.pdf works for me at least. It's on page 19.

br,
Sven

dougwilson:
Cool, that was the non loading PDF and it seems to work from my cell :) It also looks like there is a file extension you could add. Any reason not to?

svenvarkel:
I added bdoc to suffixes file. I guess that's what's needed?

svenvarkel:
My bad, added it to wrong file. I understood the structure now and added bdoc extension.

PR #32, opened 2015-07-03
Update informations for JS and PHP types

chrisyip:
- `text/javascript` should have `js` extension
- Add `text/php` for CodeMirror

### DIFF
--- a/db.json
+++ b/db.json
@@ -4298,6 +4298,10 @@
     "source": "apache",
     "extensions": ["hdf"]
   },
+  "application/x-httpd-php": {
+    "compressible": true,
+    "extensions": ["php"]
+  },
   "application/x-install-instructions": {
     "source": "apache",
     "extensions": ["install"]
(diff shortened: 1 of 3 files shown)


dougwilson:
I can take the `php` change, but unfortunately not the `text/javascript` change; we actually had to pull that out because it was causing issues downstream with our current API; we have to come up with a brand new API in a v2 before we can get it added back, sadly.

So far I just took the `application/x-httpd-php` type, as it's the only one mentioned on php.net and also works with Code Mirror, which you referenced in this PR :) Please feel free to open another PR if you want the additional and we can see if we can find a good source for them.

broofa:
Just dropping a couple breadcrumbs here:
* [the page on php.net](https://www.php.net/manual/en/install.unix.apache2.php) that mentions using `application/x-httpd-php` for Apache installations.
* [this page](https://cweiske.de/tagebuch/php-mimetype.htm) discussing PHP mime-types.

PR #39, opened 2016-01-21
Use "jpg" as the first extension of "image/jpeg"

Lanfei:
Usually, we prefer to use `jpg` as the extension instead of `jpeg`.

For Example:

```js
// ...
var contentType = req.headers['content-type']; // => 'image/jpeg'
var data = db[contentType];
if (data) {
    var ext = data['extensions'][0]; // => 'jpg'
    req.pipe(fs.createWriteStream(filename + '.' + ext));
}
// ...
```

### DIFF
--- a/db.json
+++ b/db.json
@@ -5413,7 +5413,7 @@
   "image/jpeg": {
     "source": "iana",
     "compressible": false,
-    "extensions": ["jpeg","jpg","jpe"]
+    "extensions": ["jpg","jpeg","jpe"]
   },
   "image/jpm": {
     "source": "iana"


dougwilson:
Hi! The `db.json` file is automatically build from the sources, so you cannot change it, as it will simply get reverted on the next build. Please see the contributing section https://github.com/jshttp/mime-db#contributing for where to change.

As for the ordering, I'm sure it comes from somewhere (perhaps this is the order from Apache's definitions?), so determine that would be the first step and you'll want to get the source itself changed.

This module just aggregates the given sources in the readme.

dougwilson:
Yep, you need to get https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types updated. Perhaps file an issue against the Apache project, and as soon as it's changed, we'll automatically pull it in, as we track the trunk version of Apache source.

> image/jpeg					jpeg jpg jpe

dougwilson:
It may be possible to simply specify `jpg` as the only extension in the custom types from the contributing instructions. That may bump that extension to the front.

Lanfei:
Get it, thank you for your reply.

PR #40, opened 2016-01-22
A new MIME type (application/package .pak) for Firefox OS new security module

William-Hsu:
To my knowledge, the list of Github supported MIME types is generated from the mime-db project.
We're developing a new security module on Firefox OS. I was wondering if we can add the MIME type - "application/package .pak" to the database.

Thank you. Have a nice day.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -95,6 +95,9 @@
   "application/ogg": {
     "compressible": false
   },
+  "application/package": {
+    "extensions": ["pak"]
+  },
   "application/pdf": {
     "compressible": false
   },


dougwilson:
Hi! Thanks for your pull request. The type sounds fine, though it was done to a generated file that will be replaced automatically when we pull from the Apache source again.

You should put the type in `src/custom.json`, according to the contributing section https://github.com/jshttp/mime-db#contributing

Also, if this is a new type, have you submitted it or are you considering submitting it to IANA? Instructions for that is at https://github.com/jshttp/mime-db#adding-custom-media-types .

Ideally, this module simply aggregated sources, as found in the readme https://github.com/jshttp/mime-db#mime-db . We can add new custom types here, but ideally it makes it into one of the sources we aggregate from.

William-Hsu:
Sorry for any inconvenience.
I have moved the MIME type form "src/apache-types.json" to "src/custom-types.json".

dougwilson:
Thanks for moving it!

Also, if this is a new type, have you submitted it or are you considering submitting it to IANA? Instructions for that is at https://github.com/jshttp/mime-db#adding-custom-media-types .

Ideally, this module simply aggregated sources, as found in the readme https://github.com/jshttp/mime-db#mime-db . We can add new custom types here, but ideally it makes it into one of the sources we aggregate from.

Is there any literature or specification we can link to for this type as well?

William-Hsu:
Hi, Douglas,

Oh! Now I understood what you meant.
It seems to me that I need to submit it to IANA and then the database (mime-db) will be automatically updated, right?

We are developing a new security framework on Firefox OS. So, there still are many things need to be defined. Hence, it's not right time to submit the new type to IANA. The reasons we submitted the pull request here are Github supported MIME types are generated from the mime-db project and we put all of tests on Github. ;)

I'd like to close the pull request and then submit the new type to IANA while our framework is fully developed.

Thanks for your help.
Have a nice weekend!

dougwilson:
>  It seems to me that I need to submit it to IANA and then the database (mime-db) will be automatically updated, right?

Yea, we source this module from three main places: IANA, the Apache project, and the nginx project. We also have a fourth, which is "custom", and we can add things there, but if it hasn't even been submitting to the IANA for consideration, it's kind of iffy, since ideally we just want to be an aggregation of the sources, haha.

If you have an application written up to the IANA you want to link, we can totally add it here to "custom" while it goes through the IANA process, of course :)

PR #43, opened 2016-02-06
add rng extension (Relax-NG XML)

glorieux-f:
rng is the common extension for Relax-NG in XML, http://en.wikipedia.org/wiki/RELAX_NG

* xsd is already registred (another XML schema language)
* rnc, the Relax-NG Compact syntax is also registred, by Apache (https://github.com/jshttp/mime-db/blob/master/src/apache-types.json#L337)

Correct mime-type is better for validation against schema online.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -329,7 +329,10 @@
   },
   "application/xml": {
     "compressible": true,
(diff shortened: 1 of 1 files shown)


dougwilson:
Makes sense, thanks!

PR #48, opened 2016-04-25
Add .pug extension to text/jade mime type

RockyTV:
Jade has been renamed to Pug since commit 355d3da.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -539,7 +539,7 @@
     "compressible": true
   },
   "text/jade": {
-    "extensions": ["jade"]
+    "extensions": ["jade", "pug"]
   },
   "text/javascript": {
     "compressible": true


dougwilson:
Thanks for the pull request! Looking at https://github.com/pugjs/pug I see file references changed to use `.pug`, but I wonder: did they change the MIME as well? I didn't see at a glace any references to `text/jade` there, so that's what I'm wondering. At the very least, could you dig up a link to somewhere in their documentation that specifies that they use `.pug` and what their MIME type is? That would help us fill in the now-required "sources" section for all custom types (so we don't keep ending up in situations where people wonder where these came from).

caesay:
I don't use jade/pug but from some searching in the repo, they seem to use `text/x-template` in one of their tests. This also fits with the meaning of the template mime.

https://github.com/pugjs/pug/blob/0610bdf6f58804a518322b7c002a4402e7bc0ce3/test/cases/include.script.pug

dougwilson:
So, I'm going to close this issue unless there is an actual, official word on this from their project, ideally specifically stated in their documentation (rather than guessing). Even better, they should just register the type and extensions they want to use with the IANA, which we will automatically pull in here: https://github.com/jshttp/mime-db#adding-custom-media-types

PR #50, opened 2016-08-26
Add audio/mp3

malcolmrebughini:
Because of this bug https://bugs.chromium.org/p/chromium/issues/detail?id=227004 Chromium sends `audio/mp3` as mimetype for mp3 files instead of the proper `audio/mpeg`.

Since this bug is years old, I think we might as well add the type to the list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 
 node_modules
 coverage
+
+.idea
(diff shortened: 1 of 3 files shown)


dougwilson:
lol, bugs. Thanks!

PR #51, opened 2016-10-12
Extension 'wav' lookup should return mime-type 'audio/wav' (for HTML5 audio tag)

Alcotana:
[Mime-types](https://github.com/jshttp/mime-types) npm module should return `'audio/wav'` on `mime.types['wav']` and `mime.lookup('wav')`, because html5 audio tag (Chrome, Opera, FireFox) doesn't support 'audio/wave', but 'audio/wav'
screenshot_120
screenshot_121

So I have to add the following edit to **db.json**:
screenshot_122

Now it returns `'audio/wav'`. Update please!

### DIFF
--- a/db.json
+++ b/db.json
@@ -5365,11 +5365,11 @@
   "audio/vorbis-config": {
     "source": "iana"
   },
-  "audio/wav": {
+  "audio/wave": {
     "compressible": false,
     "extensions": ["wav"]
   },
-  "audio/wave": {
+  "audio/wav": {
     "compressible": false,
     "extensions": ["wav"]
   },
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -376,11 +376,11 @@
   "audio/ogg": {
     "compressible": false
   },
-  "audio/wav": {
+  "audio/wave": {
     "compressible": false,
     "extensions": ["wav"]
   },
-  "audio/wave": {
+  "audio/wav": {
     "compressible": false,
     "extensions": ["wav"]
   },


Alcotana:
You should also add `audio/x-wav` mime type

dougwilson:
Hi and thanks for your PR! Unfortunately you cannot edit the db.json file because that is a computer-generated file and will be undone on the next build. You should keep your edits to the files in the src directory. https://github.com/jshttp/mime-db/blob/master/README.md#contributing

Alcotana:
Browsers also supports `audio/x-wav` mime type in `<audio>` tag, but for now I can only get `audio/wave`. I've made an edit to **custom-types.json** to save the changes.

dougwilson:
I think you may have accidentally added another type in this PR. Would you be willing to open another PR to keep the focus on one change at a time?

Also, I pulled your change and rebuilt the db.json file from your custom types file and I'm getting a different outcome from the db.json file you have in your PR. Basically your custom types change for wave results in the db.json file being exactly what is already in master, so there is effectively no changes being made to wave. Can you explain what you are trying to achieve? The change to custom types just reorders key, but object keys are inherently an unordered list, so we usually keep it in alphabetical order.

Alcotana:
I dropped last commit about adding another mime type.
I want [mime-types](https://github.com/jshttp/mime-types) npm module to return `audio/wav` mime type on `wav` extension lookup. Reordering elements in **db.json** somehow affects the result of the lookup. Is there are some options to make `audio/wav` default for 'wav' files (not `audio/wave` and `audio/x-wav`)?

dougwilson:
> Is there are some options to make  audio/wav  default for 'wav' files (not  audio/wave  and  audio/x-wav )?

Perhaps that is a bug or feature request for that module? This module only provides mime to extension mapping; it has no concept of extension to mime mapping, which is what you are asking about.

Alcotana:
OK, I thought reordering elements of mime types would be a hotfix to this issue.

dougwilson:
Well, the problem is that `db.json` is a computer-generated file; any hand edits will immediately be overwritten during the next library sync.

dougwilson:
It looks like nothing else came from this PR. The PR doesn't work since the `db.json` file is a generated file as noted above, so changes will immediately be overwritten.

Issue #52, opened 2016-10-12
Add 'audio/x-vorbis' mime types


Alcotana:
`audio/x-vorbis`, `audio/x-vorbis+ogg` ← `ogg`

PR #53

Alcotana:
Sorry, it is supported. Remove this issue.

dougwilson:
Hi @Alcotana I'm' not sure what happened, but your link leads to "We went looking everywhere, but couldn’t find those commits."

Martii:
@dougwilson 
Contact support as the page it pulls in references issue #51 which doesn't appear to be related. Also @Alcotana's [history](https://github.com/Alcotana?tab=overview&from=2016-09-01&to=2016-09-30&utf8=%E2%9C%93) doesn't show any PRs/commits right now like the rest of us.

Martii:
And now it shows a repo commit. heh... seems like a GH issue for support.

Alcotana:
Yeah, the link is broken because I dropped the commit from the branch. The commit got into PR #51, which is about 'wav' files. I created another PR about 'ogg': #53

dougwilson:
Closing from the conversation in #53

PR #53, opened 2016-10-14
Added "audio/x-vorbis" and "audio/x-vorbis+ogg" mime types

Alcotana:
This pull request is for the following issue:
**Add 'audio/x-vorbis' mime types** #52

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -402,6 +402,22 @@
   "audio/x-caf": {
     "compressible": false
   },
+  "audio/x-vorbis": {
+    "compressible": false,
+	"extensions": ["ogg"],
+	"sources": [
+      "https://gstreamer.freedesktop.org/data/doc/gstreamer/head/pwg/html/section-types-definitions.html",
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi @Alcotana, typically we only add types that comes from our sources (see the README). We normally do not want to add new MIME types to the custom types unless there is a very, very good reason. Can you elaborate on the following points:

1. Why should this type be added?
2. Why can this type not be registered with the IANA?
3. Why can the type not be added to the nginx or Apache projects?

The `mime-db` project is simply meant to be an aggregation of the projects listed in the README, and not a source of any types. Adding the types to the custom types is a last resort and we'll need to understand why that needs to happen.

dougwilson:
I looked into this, and the non-`x-` variants are already registered with the IANA. Why should the `x-` variants be added here? Ideally if there is a reason, the mapping should be added to either the Apache project or nginx project, I would think.

dougwilson:
Hi @Alcotana I'm going to close this PR since I never heard back on any of the questions in two months.

PR #54, opened 2016-10-14
Added "amr" extension

Alcotana:


dougwilson:
Hi @Alcotana, typically we only add types that comes from our sources (see the README). We normally do not want to add new MIME types to the custom types unless there is a very, very good reason. Can you elaborate on the following points:

1. Why should this type be added?
2. Why can this type not be registered with the IANA?
3. Why can the type not be added to the nginx or Apache projects?

The `mime-db` project is simply meant to be an aggregation of the projects listed in the README, and not a source of any types. Adding the types to the custom types is a last resort and we'll need to understand why that needs to happen.

dougwilson:
I looked into this, and it looks like `audio/amr` is already included in the Apache project, but without a file extension. It would be awesome if you can just get the `.amr` extension listed for `audio/amr` in Apache, and this project will automatically pull that in.

dougwilson:
Hi @Alcotana I'm going to close this PR since I never heard back on any of the questions in two months. Just getting the extension listed in the Apache project would be the best course of action, as it not only benefits this module, but even all users of Apache.

PR #60, opened 2016-12-31
build: support magnet extension

davesteele:
Add ".magnet" to the list of extensions for "application/x-bittorrent".

### DIFF
--- a/db.json
+++ b/db.json
@@ -4316,7 +4316,7 @@
   },
   "application/x-bittorrent": {
     "source": "apache",
-    "extensions": ["torrent"]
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks! Unfortunately we cannot accept this PR since `db.json` is an auto-generated file and your change will simply be clobbered. Please see how to contribute here: https://github.com/jshttp/mime-db#contributing

dougwilson:
In addition, when you add the entry for the extension to the proper file, please make sure you add links for where that extension is defined.

Issue #20, opened 2015-02-08
Better way to manage extension priority for multiple types


dougwilson:
We need a better way to manage extension priority for multiple types (i.e. provide an extension -> mime mapping).

dougwilson:
I am debating between two different APIs here:

  1. `require('mime-db')` will look the exact same, but there will be a new, non-enumerable property `require('mime-db').extensions` that will contain the extension map. **Pro** is backward-compatible with all the modules that depend on this library on npm. **Con** it's mixing things together.
  2. Just make `require('mime-db')` no longer be a direct mapping to the `db.json` file, but rather place it at a property (like `require('mime-db').types`). **Pro** better separation of concerns and fits ES6 modules a little better. **Con** requires a major version bump.

I'm putting this out here in case anyone has opinions on it.

/cc @Fishrock123 @jonathanong @broofa

broofa:
My $.02: Prioritize at the source level, and be explicit about which source trumps which. E.g. I'd suggest:

  1. Apache
  2. Nginx
  3. Node community submitted extensions

One thing to consider is what happens if one of these datasets (e.g. Apache) is updated to cause a conflict.  The priority here would ideally reflect the willingness of organizations to bring there dataset in line with what other orgs are doing.

dougwilson:
@broofa yes, this is what I'm doing, but in order to provide this organized list for extension -> mime lookup (we already do this for mime -> data), we will have to actually provide the list in some way. Right now the modules on top of this are simply iterating over the db and just building the list, but of course that means their list is going to be organized alphabetically by mime type, rather than order by source preference (this is an issue specific to extension -> mime lookups).

What I'm currently contemplating is what the API of `require('mime-db')` should look like when providing this second mapping.

jonathanong:
@dougwilson something like

```js
// require() style
import types from 'mime-db/types.json'
import extensions from 'mime-db/extensions.json

// all at once
import { types, extensions } from 'mime-db'

// by default
import mime from 'mime-db'
mime.extensions
mime.types
```

jonathanong:
i don't mind breaking changes :D

broofa:
I'm confused about what the issue is here.

> extension priority for multiple types

[...]

PR #61, opened 2017-01-11
Adds application/geo+json custom type.

TobiTonner:


dougwilson:
I spoke a bit too soon. Looks like due to the way `mime-types` and other libraries are trying to create a reverse map from this module, the inclusion of `.json` is causing those modules to think `.json` equals `application/geo+json` with this patch.

Pending #20 for now I'm going to merge with altered to remove the `.json` extension and leave only `.geojson`.

PR #65, opened 2017-03-04
Adds iWork Keynote, Pages and Numbers

jonthomp:
Adds the types and extensions for Apples iWork Keynote, Pages and Numbers applications.

They don't appear to be registered with the IANA and I see from some previous PRs that ideally, we would look into having these added to the Apache or NGINX projects? Can you point me in their direction and I'll get the ball rolling but also accept this PR as short term solution for something I'm working on?

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -699,5 +699,17 @@
     "sources": [
       "https://developer.mozilla.org/en-US/docs/Web/WebGL/Adding_2D_content_to_a_WebGL_context"
     ]
+  },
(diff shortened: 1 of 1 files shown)


dougwilson:
Just submit a patch or issue to their issue tracker.

dougwilson:
Hi @jonthomp I'm going to close this since I never heard back and couldn't find any references myself. Please feel free to ping me if you have the references and I can re-open or you can re-make the PR with the references :) !

PR #67, opened 2017-03-31
official mime type for markdown

captain-digitalsailors:
as per RFC 7763 - https://tools.ietf.org/html/rfc7763

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -582,6 +582,10 @@
   "text/less": {
     "extensions": ["less"]
   },
+  "text/markdown": {
+    "compressible": true,
+    "extensions": ["markdown","md"]
+  },
   "text/n3": {
     "compressible": true,
     "sources": [


KyleAMathews:
Any chance these breaking changes could be done as major releases? Gatsby was relying on `text/x-markdown` so this change broke things for us https://github.com/gatsbyjs/gatsby-starter-blog/issues/53

dougwilson:
If this was brought up 5 months ago maybe, but that ship has sailed. Regardles, our contract is on our API, not the database content which can change at any time, inaccordance to IANA registrations.

dougwilson:
Sorry 3 months ago. We had the change pend for 2 months to give time for feedback which none came. Please subscribe to this repository if you want to keep an eye on changes or pin to the minir version if you want to vet any changes to the database before you get an updated database.

KyleAMathews:
Fair enough :-) I'll start watching so I can make sure Gatsby is updated for mime changes before they're pushed out.

Thanks! This is an awesome resource!

dougwilson:
No problem, and sorry if I came off harsh at all; was just typing up from my phone :) There isn't much activity in here and I try to be careful, but being a database is outside the scope of semver unfortunately, so under semver one can argue that simply updating the database is a patch release, but every database update has to update something, haha. We can either be at 1.0.30, 1.30.0, or 30.0.0 depending on how the database updates are seen, though then if we change the actual DB format, if every data update was a major, you'll never be able to tell if the way code interfaces with the database is why the major bump or just a routine data update, lol.

KyleAMathews:
Haha no problem at all :-) And totally agree, semver-ing on database changes would be meaningless as pretty much every update here is a data update.

PR #69, opened 2017-04-06
Add virtual disk image types

jhermsmeier:
This adds a range of commonly used virtual disk image types.


jhermsmeier:
@dougwilson yup, added sources now.

PR #71, opened 2017-04-11
add audio/aac mime

taina0407:


dougwilson:
H @taina0407 , the `db.json` file cannot be edited, as it is a generated file from the files in the `src/` directory. Please update the PR following the instructions in https://github.com/jshttp/mime-db#contributing

PR #72, opened 2017-04-11
Add audio/aac and audio/aacp types

taina0407:


dougwilson:
P.S. There is a lot of text on that source, and I tried to read through it, but may have missed it, so if it's there, that's OK. Let me know and perhaps we can even update (or add as an additional source link) to point to the specific section in the page.

dougwilson:
Closing since I never heard back.

PR #74, opened 2017-05-10
Add application/dxf to custom-types.json

Hobbs1210:
Add application/dxf to custom-types.json

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -25,7 +25,10 @@
   },
   "application/dxf": {
     "compressible": false,
-    "extensions": ["dxf"]
+    "extensions": ["dxf"],
+	"sources": [
+      "http://filext.com/file-extension/DXF"
+    ]
   },
   "application/ecmascript": {
     "compressible": true
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -23,6 +23,10 @@
       "https://www.iana.org/assignments/media-types/application/dash+xml"
     ]
   },
+  "application/dxf": {
+    "compressible": false,
+	"extensions": ["dxf"]
+  },
   "application/ecmascript": {
     "compressible": true
   },
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -26,8 +26,8 @@
   "application/dxf": {
     "compressible": false,
     "extensions": ["dxf"],
-	"sources": [
-      "http://filext.com/file-extension/DXF"
+    "sources": [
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks for adding in that link! I'm not sure if I'm missing it, but I don't see anything in there saying that the MIME type is "application/dxf". Which page is that on?

Martii:
> ... saying that the MIME type is "application/dxf" ...

dxf spotlight

Martii:
Do seem to have a collision with additional checks at http://www.fileformat.info/info/mimetype/image/vnd.dxf/index.htm and https://www.iana.org/assignments/media-types/media-types.xhtml using `image/vnd.dxf` as the registered primary though.

dougwilson:
So then should the MIME be `image/vnd.dxf`? The fileformat.info is not a primary source, and anyone can add anything there, so ideally if there is some information from the folks to actually control the `.dxf` extension, that would clarify what the MIME should be.

Martii:
> So then should the MIME be `image/vnd.dxf`?

I think so. It would be neat to have an alias field set at some point if it doesn't already exist. Some types started out with one type and then ended up being registered as another from back in at least my day... this even happens currently during development.

> ... that would clarify what the MIME should be.

[.../blob/`6f42b46`/src/apache-types.json#L2838-L2840](https://github.com/jshttp/mime-db/blob/6f42b46533c291ae4d390691388eee0e11bf29c8/src/apache-types.json#L2838-L2840) is where it currently is already in the DB. Jodi Moline is the registration contact although emails/contact info can change over the years *(decades in this case since it was 1996)*.

Hobbs1210:
DXF's are really inconstant since they are used by a variety of apps. i have found a MIME list on Harverd.edu that includes application/dxf.
http://hul.harvard.edu/ois/////systems/wax/wax-public-help/mimetypes.htm

dougwilson:
@Hobbs1210 but once again, that is just a generic list. If I merge this, then that just contributes to the issue of what is the real MIME. As far as I can tell, the MIME `image/vnd.dxf` is registered with the IANA, which is the official source of real MIMEs. The `application/dxf` is not registered. At this point, I suggest that either (a) get it registered with IANA and we'll automatically pull it in or (b) get a primary source that says `.dxf` = `application/dxf`.

Hobbs1210:
That's not a problem. I will close the pull request since image/vnd.dxf can be used for dxf files.

PR #75, opened 2017-05-10
ANSI art and ARJ archive formats

NuSkooler:
This PR contains `application/x-arj` ARJ archive and `text/x-ansi` ANSI art MIME types.

### DIFF
--- a/db.json
+++ b/db.json
@@ -4613,7 +4613,6 @@
   },
   "application/x-lzh-compressed": {
     "source": "apache",
-    "compressible": false,
     "extensions": ["lzh","lha"]
   },
   "application/x-makeself": {
--- a/db.json
+++ b/db.json
@@ -4364,6 +4364,10 @@
     "source": "apache",
     "extensions": ["dmg"]
   },
+  "application/x-arj": {
+    "compressible": false,
+    "extensions": ["arj"]
+  },
   "application/x-authorware-bin": {
     "source": "apache",
     "extensions": ["aab","x32","u32","vox"]
@@ -6368,6 +6372,10 @@
     "charset": "UTF-8",
     "compressible": true,
     "extensions": ["vtt"]
+  },
+  "text/x-ansi": {
+    "compressible": true,
+    "extensions": ["ans"]
   },
   "text/x-asm": {
     "source": "apache",
--- a/db.json
+++ b/db.json
@@ -4613,6 +4613,7 @@
   },
   "application/x-lzh-compressed": {
     "source": "apache",
+    "compressible": false,
     "extensions": ["lzh","lha"]
   },
   "application/x-makeself": {
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -633,6 +633,15 @@
        "http://tools.ietf.org/html/draft-pantos-http-live-streaming-13"
     ]
   },
+  "text/x-ansi": {
+    "compressible": true,
+    "extensions": ["ans"],
+    "notes": "ANSI art",
+    "sources": [
+      "https://lists.freedesktop.org/archives/xdg/2006-March/006214.html",
+      "https://de.wikipedia.org/wiki/ANSI-Art"
+    ]
+  },
   "text/x-gwt-rpc": {
     "compressible": true
   },
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -654,6 +654,15 @@
   "text/x-lua": {
     "extensions": ["lua"]
   },
+  "application/x-lzh-compressed": {
+    "compressible": false,
+    "extensions": ["lzh","lha"],
+    "sources": [
+      "https://en.wikipedia.org/wiki/LHA_(file_format)",
+      "https://www.backblaze.com/b2/docs/content-types.html",
+      "http://www.help.websiteos.com/mime_types.phtml"
+    ]
+  },
   "text/x-markdown": {
     "compressible": true,
     "extensions": ["markdown","md","mkd"]
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -654,15 +654,6 @@
   "text/x-lua": {
     "extensions": ["lua"]
   },
-  "application/x-lzh-compressed": {
-    "compressible": false,
-    "extensions": ["lzh","lha"],
-    "sources": [
-      "https://en.wikipedia.org/wiki/LHA_(file_format)",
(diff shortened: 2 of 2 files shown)


src/custom-types.json
```diff
@@ -236,6 +236,14 @@
   "application/x-7z-compressed": {
     "compressible": false
   },
+  "application/x-arj": {
+    "compressible": false,
+    "extensions": ["arj"],
+    "sources": [
+      "https://en.wikipedia.org/wiki/List_of_archive_formats",
+      "http://www.wussu.com/various/mimetype.htm"
```
dougwilson: When you get a chance, can you add a link to where this is defined, please note that

> This needs to be a [primary source](https://en.wikipedia.org/wiki/Primary_source);
  links to type aggregating sites and Wikipedia are _not acceptible_.

NuSkooler: Apologies on the Wikipedia link. I missed that note.

I did some digging last night, but have not come up (yet) with an official source. An attempt to email the author (Robert Jung, http://www.arjsoftware.com/) resulted in an email bounce.

A few notes that may be of interest that lead me to `application/x-arj` vs some of the others that have floated around over the years:
* [file(1) uses it](https://opensource.apple.com/source/file/file-37/file/magic/Magdir/archive)
* Most Linux desktop relations seem to use this (for icons, launchers, etc. integration)
* Various references to mime types in older documentation (e.g. the second link in sources I provided is from 1998)

Some issues with the above: 
`application/arj` is also referenced here and there commonly. For example, https://goo.gl/Wl1XoQ (1999). 

I did find a very old email reference for Robert in an original ARJ distribution from the 90's. I have mailed him & hope to hear back (hasn't bounced back yet!)

Thanks!

Martii: > `application/arj` is also referenced here and there commonly. For example, https://goo.gl/Wl1XoQ (1999).

I would guess :-1:  to this source. Has several that are incorrect... is there a revised edition book of this?

> [file(1) uses it](https://opensource.apple.com/source/file/file-37/file/magic/Magdir/archive)

Leaning towards :-1:  since that's in an archive that isn't current with apples opensource e.g. try the parent folder and try to find it as an actively referenced path in their folder view.

> Most Linux desktop relations seem to use this (for icons, launchers, etc. integration)

Confirmed with `application/x-arj` in my distro.

> lead me to `application/x-arj`

If I were to take an educated guess this is probably correct... since I can't find a registration and being of that time period. :wink:

> ... hope to hear back ...

Wish the best of luck here, seriously because I went on a hunt too with little success. :smile: ... older MIME/Content-Types are difficult to validate from that time period as it was relatively new and documentation was sparse since it was in print primarily back in the dawn of computer internet time where gophers roamed freely.

dougwilson: Thanks for the research done so far. I can still go forward with merging if we can't find a definitive source, but I want to make sure that we did due-diligence before adding in types that don't have any real concrete source.

If there is ample evidence that both are necessary, we could even add both.

Martii: Part of the reason why arj didn't make it as a popular web compression method is the same as LZH where it wasn't parse-able without downloading the entire archive. PK *(zip)* is and has been to my knowledge since the olden days. Similar to how ID3 v1.0/1.1 data blocks for mp3's is at the tail end of the binaries vs v2.x which is at the forefront.

I even tried it on a clean Apache with zero mods and it comes with no `Content-Type` header on a response. Not sure if I have my IIS available from that time period of MSDN to see if anything was defaulted but I doubt it.

db.json
```diff
@@ -6369,6 +6373,10 @@
     "compressible": true,
     "extensions": ["vtt"]
   },
+  "text/x-ansi": {
+    "compressible": true,
+    "extensions": ["ans"]
+  },
```
Martii: Can't seem to find any original source what-so-ever for this one for MIME/Content-Type but I have seen these in archive files back in the day. Typically just about any file extension *(recall `.nfo` mainly back in my BBS days with X/Y/Z-Modem)* could have 8-bit ANSI escaped sequences in them. 7-bit is pretty bland but can have ascii art in them too. Mac vs IBM/PC is also a factor with ANSI since it wasn't industry standard and rivals. Thus the birth of Unicode came along... too unify character sets. Anyhow... history lesson from my perspective complete. ;)

The real question I think is... is there software out there that needs a MIME type for these... if so then probably the developer version of `text/x-ansi` versus a production/release/golden/official is a good bet.

NuSkooler: I see someone else remembers the BBS days :) My best (only) references for `text/x-ansi`:
* https://lists.freedesktop.org/archives/xdg/2006-March/006214.html
* https://de.wikipedia.org/wiki/ANSI-Art (references the first)
* https://github.com/eight04/ansi-viewer

For my purposes, I've written a "modern" (Ha!) BBS in Node.js that relies on node-mime, which in turn relies on mime-db for various mappings.

src/custom-types.json
NuSkooler: Thanks everyone for looking into these! I'll give it a few days before I give up on the email to Robert Jung. I think I'll ask in the BBS networks as well (yes, they _still_ exist!).

[Modern Linux file(1)](https://github.com/file/file/blob/daaf6d44768677aca17af780bba0a451fbb69ac8/magic/Magdir/archive) seems to also state `application/x-arj`

db.json
Martii: That's probably why I was gravitated towards this PR as it sparked some curiosity and old neurons. ;) Definitely miss my BBS days.

src/custom-types.json
NuSkooler: Quick update:
I've asked around on FidoNet: Some responses, but no luck on an official/primary source. No response from Robert Jung, either :(

NuSkooler: Unfortunately I may have to give up on finding an official source on this. No emails back, no responses on FTN networks with good info, etc. 

Where does that put us?

db.json
NuSkooler: Are the above "enough" or is this one a dead end?

src/custom-types.json
dougwilson: Sorry, I didn't merge this PR since it had two types with issues. The ARJ one was resolved, but the ANS one seems less likely to be. To that end, I am going to merge just the ARJ part of the PR and the ANS change can be re-opened as a new PR if it is desired to work through.

NuSkooler: @dougwilson Sounds great, thank you!

PR #82, opened 2017-06-21
add an orgmode mimetype

jkitchin:
This adds a new mimetype for orgmode files (http://orgmode.org).

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -652,6 +652,10 @@
     "compressible": true,
     "extensions": ["mkd"]
   },
+  "text/x-orgmode": {
+    "compressible": true,
+    "extensions": ["org"]
+  },
   "text/x-processing": {
     "compressible": true,
     "extensions": ["pde"],
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -652,9 +652,13 @@
     "compressible": true,
     "extensions": ["mkd"]
   },
-  "text/x-orgmode": {
+  "text/x-org": {
     "compressible": true,
-    "extensions": ["org"]
+    "extensions": ["org"],
+    "sources": [
+      "https://lists.gnu.org/archive/html/emacs-orgmode/2011-01/msg00971.html",
+      "http://orgmode.org"
+    ]
   },
   "text/x-processing": {
     "compressible": true,


jkitchin:
I have added a link to an email discussing this mimetype, and to the orgmode website. There doesn't seem to be anything listed in IANA (https://www.iana.org/assignments/media-types/media-types.xhtml#text).

dougwilson:
At the very least, that email you linked to ends with:

> ...which doesn't necessarily mean it's correct. I have no idea myself,
so I'm cc-ing the Org list, perhaps somebody there can provide more
insight.

So if you can link to the response where the orgmode project confirmed the mime association, that would be good enough as well.

jkitchin:
The official mimetype of org files is now "The de facto mimetype for org attachments is text/x-org." and is documented at http://orgmode.org/index.html.

jkitchin:
Apparently the x- prefix is being deprecated, so the org-mode project now lists the mimetype as "The de facto mimetype for org attachments is text/org".

josephmturner:
Hello @dougwilson @jkitchin !

> So if you can link to the response where the orgmode project confirmed the mime association, that would be good enough as well.

According to [the Org mode website](https://orgmode.org/#more), "The de facto mimetype for Org files is text/org. Org files use the .org extension." Is this good enough for a PR like this one to be merged?

Thank you!

wesleytodd:
Hey @josephmturner. I would suggest opening a new PR. If you could also include those references inline into the description. Ideally if you could also get someone to describe what is implied by "defacto" that would be awesome.

josephmturner:
@wesleytodd  Thanks for the suggestions.  I realize now that `mime-db` already contains an [`.org` => `text/x-org` mapping](https://github.com/jshttp/mime-db/blob/master/db.json#L8094).

This mapping is already being ignored in [mime-types](https://github.com/jshttp/mime-types) since it has lower preference than the IANA `.org`=> `application/vnd.lotus-organizer` MIME type.  So I think that adding an `.org` => `text/org` would not meet my need, which is for [mime](https://github.com/broofa/mime) or [mime-types](https://github.com/jshttp/mime-types) to do either of these:

```
mime.lookup('org') // 'text/org' ('text/x-org' would be okay also)
```

PR #84, opened 2017-06-27
Add vue.js component file type

dickeylth:


src/custom-types.json
```diff
@@ -583,6 +583,13 @@
       "http://facebook.github.io/react/docs/getting-started.html"
     ]
   },
+  "text/vue": {
+    "compressible": true,
+    "extensions": ["vue"],
+    "sources": [
+      "http://vuejs.org/v2/guide/single-file-components.html"
```
dougwilson: Please provide a primary source for the mapping of the mime type `text/vue` to the extension `.vue`. I don't see the mime type anywhere on that page.

dickeylth: I'm wondering if that's really necessary? There's nothing about `text/jsx` on the page <http://facebook.github.io/react/docs/getting-started.html> either.

dougwilson: Yes, it is necessary. Otherwise, you need to follow https://github.com/jshttp/mime-db#adding-custom-media-types

dougwilson: > There's nothing about text/jsx on the page http://facebook.github.io/react/docs/getting-started.html either.
 
It looks like their site has been re-arranged. You can see that it now is redirecting to `hello-world.html`. Here is what the page used to look like: https://shripadk.github.io/react/docs/getting-started.html

dougwilson:
Sorry, that last review was for another PR, but still looking for the page that states the mime type for this one.

yyx990803:
This is what `*.vue` files are: http://vue-loader.vuejs.org/en/start/spec.html (I am the author of Vue)

Technically they are only used as a source for compilation, but since JSX files also has a mime type, I guess `text/vue` can be added too.

dougwilson:
This module is just a list of the types, not a registration service. If you want to declare a MIME type, you need to follow the IANA instructions at http://tools.ietf.org/html/rfc6838#section-5 .

yyx990803:
Then what is `custom-types.json` for? AFAIK jsx is not registered with IANA.

dougwilson:
Yes, things were accepted in the past that shouldn't have been. It's hard to change that. This is only supposed to be a collection of actual, registered extensions / mimes, and people havestarted just treating this like they can just list here whatever they feel like. This is not a registration service. The custom types was originally created for historical types, before things like the IANA process existed.

If you get the community review process completed, we can add it to custom types while it works it's way through to the IANA list, but you need to at least start the IANA process.

dougwilson:
I wil try to expand the https://github.com/jshttp/mime-db#adding-custom-media-types in the README, and if you have any tip on how to make it clearer on how to get new types added, always feel free to make a PR to the README as well!

dougwilson:
Closing due to no response. You can add the type back if the Vue.js folks either (a) register the type with the IANA or at the very least (b) list the type on their website.

PR #85, opened 2017-06-27
Add application/fido.trusted-apps+json file type

kspearrin:
For FIDO U2F trusted facet JSON files.

> The response MUST set a MIME Content-Type of "application/fido.trusted-apps+json".

Read more: https://fidoalliance.org/specs/fido-u2f-v1.0-ps-20141009/fido-appid-and-facets-ps-20141009.html

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -33,7 +33,6 @@
     "compressible": false
   },
   "application/fido.trusted-apps+json": {
-    "extensions": ["fidou2f"],
     "sources": [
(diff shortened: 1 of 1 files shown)


kspearrin:
I quoted it from the page in the original description above: 

> The response MUST set a MIME Content-Type of "application/fido.trusted-apps+json".

There is no specification on what the extension should be, only that the mime type be `application/fido.trusted-apps+json`. It could be any extension.

I assume that for the purposes of GitHub pages serving the correct mime type though, some type of unique extension is required.

dougwilson:
Yea, I saw the mime type. So either remove the file extension from the PR, provide the source for where the extension is specified, or go through the IANA process to specify an extension.

PR #88, opened 2017-08-08
Add mjs to application/javascript

bmeck:


src/custom-types.json
```diff
@@ -78,7 +78,11 @@
   },
   "application/javascript": {
     "charset": "UTF-8",
-    "compressible": true
+    "compressible": true,
+    "extensions": ["mjs"],
+    "sources": [
+      "https://github.com/nodejs/node-eps/blob/master/002-es-modules.md"
```
dougwilson: I don't see the mime listed on that page, though there is a link to another page saying .mjs should be type application/javascript+module

Can the page be updated to clarify what the mime is or link to a different source that does?

bmeck: It links to https://github.com/tc39/ecma262/issues/322 which rejects the new subtype of `+module` and can be cross referenced in https://github.com/whatwg/html/issues/558 and step 7 of https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-single-module-script . I will write up a PR and merge it for the node EP explicitly stating that it needs to be the same.

bmeck: https://github.com/nodejs/node-eps/pull/61 is approved pending confirmation of what the preferred MIME type should be. Either `text/javascript` or `application/javascript`.

dougwilson: Yea, unless it comes from one of our sources, it's standard procedure to collect this before adding to custom. It seems to have been useful, as it sounds like the act of clarifying may have changed what it is going to be. If it does end up landing as `text/javascript`, please feel free to update the PR and I can merge 👍

bmeck: removed any preference since standards disagree, any web compatible MIME is the main need so this looks fine.

bmeck:
Addendum that this has been incorporated into an Internet Draft for being merged into the IANA MIME registry : https://datatracker.ietf.org/submit/status/88872/ . That may take a bit of time to land however.

bmeck:
<3 , will be watching for release since this is an upstream dep for lots of things

dougwilson:
Alright, 1.30.0 is being published with the change now 🤞

PR #97, opened 2017-10-16
Add application/wasm

lukewagner:
Eventually, 'application/wasm' should be officially registered.  In the meantime, it would be very nice to have this working on Pages since [certain WebAssembly JS APIs](https://github.com/WebAssembly/design/blob/master/Web.md#additional-web-embedding-api) require the application/wasm MIME type to work at all.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -262,7 +262,8 @@
     "extensions": ["wasm"],
     "notes": "WebAssembly module",
     "sources": [
-      "https://webassembly.github.io/spec"
+      "https://webassembly.github.io/spec",
+      "https://github.com/WebAssembly/design/blob/master/Web.md"
     ]
   },
   "application/x-7z-compressed": {
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -262,8 +262,7 @@
     "extensions": ["wasm"],
     "notes": "WebAssembly module",
     "sources": [
-      "https://webassembly.github.io/spec",
-      "https://github.com/WebAssembly/design/blob/master/Web.md"
+      "https://webassembly.github.io/spec/binary/conventions.html"
     ]
   },
   "application/x-7z-compressed": {
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -257,6 +257,14 @@
       "http://www.iana.org/assignments/media-types/application/vnd.sun.wadl+xml"
     ]
   },
+  "application/wasm": {
+    "compressible": true,
+    "extensions": ["wasm"],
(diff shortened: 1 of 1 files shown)


src/custom-types.json
```diff
@@ -257,6 +257,14 @@
       "http://www.iana.org/assignments/media-types/application/vnd.sun.wadl+xml"
     ]
   },
+  "application/wasm": {
+    "compressible": true,
+    "extensions": ["wasm"],
+    "notes": "WebAssembly module",
+    "sources": [
+      "https://webassembly.github.io/spec"
```
dougwilson: Yea, definitely can get it added. Can you add a link to a page that at least mentions the file extension `.wasm` and the MIME type `application/wasm`? The page here is long, but skimming through it I didn't see either mentioned.

lukewagner:
You bet!  PTAL

dougwilson:
Thanks! I just took a look, but I see no mention of the file extension `.wasm` on the page. Where is it? I did find the MIME through ctrl+f searching. The page is long and so maybe I'm just not able to figure out where on the page the file extension is.

dougwilson:
P.S. I see reading on there it talks about reading the `application/wasm` MIME as a byte array. Are you sure that is compressible? Binary doesn't mean it's not compressible, just double-checking.

lukewagner:
Oh sorry!  We mention the canonical file extension in the core spec here:
    https://webassembly.github.io/spec/binary/conventions.html

But we don't mention the canonical MIME type in the same core spec.  I'll file and issue because probably this is a good thing to say right next to the recommended extension.

While wasm is a binary format, it is explicitly designed to leave compression to a higher level so compression is a large win.  For example, in the [canonical demo](http://webassembly.org/demo), the 12mb .wasm file gzips down to 3.8mb.

lukewagner:
(Updated PR to link directly to https://webassembly.github.io/spec/binary/conventions.html which avoids searching through the whole spec and also should soon mention 'application/wasm' MIME type.)

lukewagner:
Ok, https://webassembly.github.io/spec/binary/conventions.html now mentions both .wasm and application/html in the same paragraph.  Anything else needed to merge this PR?

dougwilson:
Merged 👍

lukewagner:
Thanks!  Do you happen to know how often this repo is uplifted into GitHub Pages?

dougwilson:
Hi @lukewagner I'm not 100% sure. I know GitHub Pages uses this project to get the types, but I'm not familiar with how that gets updated, and if it is just time based or is dependent on something happening here.

PR #99, opened 2017-10-31
+ text/shex

ericprud:
It's been submitted to IANA but these things take a long time. I'll clean stuff up when it's finally integrated.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -756,6 +756,12 @@
     "compressible": true,
     "extensions": ["rtf"]
   },
+  "text/shex": {
+    "extensions": ["shex","shexc"],
+    "sources": [
+       "http://shex.io/shex-semantics/"
+    ]
+  },
   "text/slim": {
     "extensions": ["slim","slm"],
     "sources": [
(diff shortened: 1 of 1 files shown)


dougwilson:
Nice! Sorry I missed your PR before, it's no problem to add early when there is a reference source. I looked through it and everything checks out except I don't see any mention at all about the `.shexc` file extension. Can you point me to where that file extension is mentioned (probably easy to miss)? Thanks!

ericprud:
The `.shexc` extension is a reaction to apparent practice. ShEx has three dialects:
1. ShExC (compact syntax) - what humans want to read and write.
2. ShExJ (JSON-LD) - the abstract syntax rendered in JSON.
3. ShExR (RDF) - the RDF interpretation of that JSON-LD.
I originally had in mind that folks `.shex` and `.jsonld` for 1 and 2 but have noticed folks using `.shexc` sometimes. Maybe the best thing to do is to hold a hard line and say the extension is only `.shex` in which case getting the wrong media type for a `.shexc` file nudges them in the direction of conformance. I guess I'm happy either way.

PR #100, opened 2017-11-03
Moved the mimetype to the proper file.

kmwarter:


src/custom-suffix.json
```diff
@@ -2,5 +2,8 @@
   "+json": {
     "compressible": true,
     "notes": "+json means it is JSON, which is compressible."
-  }
+  },
+  "text/x-python-script": {
```
dougwilson: So this doesn't work with the `build` script. This file is only for suffixes (see the `README.md` file in the `src/` directory for what each file should contain. Merging this won't actually add the type at all in the current file.

dougwilson:
Closed since I haven't heard back. Please feel free to open a new PR when you make the requested changes :) !

PR #105, opened 2017-11-26
Update hjson mime type to application/hjson

barraq:
# Description

The `hjson` [RFC](https://hjson.org/rfc.html#rfc.section.1.3) defines the mime type to be
`application/hjson` not `text/hjson`; see https://hjson.org/rfc.html#rfc.section.1.3

This PR update the mime type according to the RFC

### DIFF
--- a/db.json
+++ b/db.json
@@ -428,6 +428,9 @@
   "application/held+xml": {
     "source": "iana"
   },
+  "application/hjson": {
+    "extensions": ["hjson"]
+  },
   "application/http": {
     "source": "iana"
   },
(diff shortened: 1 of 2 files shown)


barraq:
@dougwilson thanks for you review, I sorted the key and added the source. When you say that "it came from node-mime", does it mean I should update this package too? Initially I was planning on creating a PR to this package but then I saw https://github.com/broofa/node-mime/blob/master/src/build.js#L9, isn't `node-mime` using `mime-db`?

dougwilson:
No, because it uses this package, but that was not the case before. When that module started usibg this module, we imported all the types from that module here, which is where the entry came from in the first place (I was just noting why there was no source before). We need the sources goibg forward so we can tell where they came from when questions come up, haha.

PR #107, opened 2017-11-29
resolves #106 add mime type for AsciiDoc

### LINKED ISSUE #106: add mime type for AsciiDoc files
Add the mime type `text/asciidoc` for AsciiDoc files. AsciiDoc files commonly end with `.adoc`, though the extensions `.asciidoc`, `.asc`, and sometimes `.ad` are also recognized. If we have to pick only one, it should be `.adoc`.

The mime type text/asciidoc follows the pattern used for Markdown, a very similar language. (see http://www.iana.org/assignments/media-types/text/markdown).

mojavelinux:


mojavelinux:
There is no source for this mime type as far as I know. Do you want just want a reference to the language itself?

dougwilson:
This is not a registry, just an aggregate of existing sources. Types without a concrete source can't be accepted.

mojavelinux:
Isn't that what "custom" is for?

mojavelinux:
It's absurd that there's no mime type for AsciiDoc. It's over 10 years old and well accepted on the internet.

mojavelinux:
Btw, I am the lead maintainer of the AsciiDoc language.

dougwilson:
https://github.com/jshttp/mime-db/blob/master/README.md#adding-custom-media-types

dougwilson:
At somepoint in the 10 years the maintainers of AsciiDoc could have easily registered the type and extensions with the IANA. Just register it and we'll pull it in, as noted in our README, no worries.

mojavelinux:
Can you please explain to me what the use of the term "custom" is in this context then?

I don't know how this mime type never got registered (and Markdown did), but I as the current maintainer, I suppose I can take on the responsibility of doing so.

mojavelinux:
In other words, using the term custom here is misleading.

dougwilson:
idk i am not the original author. I know as much behind the reason for the name as you know why the type was never registered it seems.

dougwilson:
Ideally we should make https://github.com/jshttp/mime-db/blob/master/README.md#adding-custom-media-types clearer. How can the adding custom media types readme section be made clearer for how to add custom media types? I'm not trying to be mean; I don't know what people are expecting to see in the readme to know how to proceed so any feedback on how to improve our documentaion on what you should do is most welcome.  Either as just feedback here or a PR with docs changes, doesn't matter.

dougwilson:
The biggest issue is that since this module is not a registry, it doesn't have governance, type conflict resoluton or guidelines. We didn't set any of that up because the IANA has done that many years ago. We needed programmatic access to the contents and so made this. There are scripts in here that HTTP fetch from the sources that are listed in the README and put them into the JSON format we needed and thought others may find it useful.

Of course all of this is volunteer work and we don't have the time or money available to be an actual registry, so just want to simply aggregate. There are local overrides from when another module was merged into this one ("node-mime") or we usually will still add types that are not in IANA reluctantly as long as a link to a full specification is included that states the mime and extensions. We have rejected many PRs that don't incldue this because we're not a registry, so this shouldn't be the first place a type is "registered". I hope that makes sense.

Thoughts on improving the readme are very welcome (we work mainly on people deciding to contribute little bits of their time here and there as things like PRs).

I hope you are not mad, becaude that isn't my intention, but we have to draw lines somewhere and it really sucks of we didn't explain well upfront what they were before you invested valuable time, and I'm sorry.

mojavelinux:
I'm not mad, and certainly not mad at you. I'm just annoyed by the bureaucracy. As the person who can speak with authority about the language, I would expect that would have been enough. But if this library adheres strictly to the authority of IANA, I can respect that. I've initiated the effort to get the mime type for AsciiDoc registered with IANA.

I'll give the wording some thought and propose a term which I think makes this policy more clear. To me, "custom" clearly means that custom types are permitted here. But I need a few moments to think about what the right term should be.

dougwilson:
I mean, all we need is a concrete specification to link to. If you publish it as the authority of AsciiDoc on whatever the website is for that ... that's good enough, because that's all I'm even looking for.

> To me, "custom" clearly means that custom types are permitted here.

I can see that, but at the same time, the README as a section specially titled "Adding custom media types" and explains what do to (which is not what you did.. and I just don't understand how that came to be). Is that not enough? I don't understand what lead to what is all, and adding clarification to the README would help clarify not just this but also help future people who may also run into the same logic here.

mojavelinux:
> I mean, all we need is a concrete specification to link to. If you publish it as the authority of AsciiDoc on whatever the website is for that ... that's good enough, because that's all I'm even looking for.

:+1:

mojavelinux:
> I can see that, but at the same time, the README as a section specially titled "Adding custom media types" and explains what do to

The problem is, "custom" has an extremely strong and well-defined meaning which is "defined here, not somewhere else". I propose the term "supplemental" instead, which is taken to mean that something upstream is deficient and needs to be patched. But it does not imply that something new is being introduced, or that this is the authority.

PR #108, opened 2017-12-21
WOFF is compressible

evilebottnawi:
See https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization?hl=en

> Note: Consider using Zopfli compression for the EOT, TTF, and WOFF formats. Zopfli is a zlib compatible compressor that delivers ~5% file-size reduction over gzip.

### DIFF
--- a/db.json
+++ b/db.json
@@ -384,7 +384,7 @@
   },
   "application/font-woff": {
     "source": "iana",
-    "compressible": false,
+    "compressible": true,
     "extensions": ["woff"]
   },
   "application/framework-attributes+xml": {


dougwilson:
Thanks for your pull request! Unfortunately we cannot accept it as this modifies db.json which is a generated file and will be overwritten automatically.

If you are interested in making the change, please feel free to open a new PR following our contribution instructions: https://github.com/jshttp/mime-db/blob/master/README.md#contributing :+1:

PR #109, opened 2017-12-22
fix: `woff` is compressible

evilebottnawi:
See https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization?hl=en

> Note: Consider using Zopfli compression for the EOT, TTF, and WOFF formats. Zopfli is a zlib compatible compressor that delivers ~5% file-size reduction over gzip.

Based on https://github.com/jshttp/mime-db/pull/108.
@dougwilson friendly ping :+1:

### DIFF
--- a/db.json
+++ b/db.json
@@ -384,7 +384,7 @@
   },
   "application/font-woff": {
     "source": "iana",
-    "compressible": false,
+    "compressible": true,
     "extensions": ["woff"]
   },
   "application/framework-attributes+xml": {
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -38,7 +38,7 @@
     ]
   },
   "application/font-woff": {
-    "compressible": false,
+    "compressible": true,
     "sources": [
       "https://github.com/h5bp/server-configs-apache/issues/42"
     ],


dougwilson:
Thanks @evilebottnawi ! I didn't know this before, but I see it was explicitly set to `false`, but you are changing it to `true`. I would have asked you to add to the `sources` array a link explaining why they can be compressed, but I see it already contains a link that explains why they shouldn't be compressed. I'm no longer certain what the answer is supposed to be.

I don't want to just change it from `false` to `true` when it looks like someone went through a lot of effort explaining why it should be `false` and there is no detailed counter answer for why it should be `true` that addresses the content in the link we already have.

What would be the next step here to help clarify?

alrra:
>  Consider using Zopfli compression for the... WOFF formats.

That actually refers to using Zopfli as the compression algorithm when creating the WOFF file (using tools such as [this one](https://github.com/bramstein/sfnt2woff-zopfli)), so basically using Zopfli "internally".

Compressing an existing WOFF file is similar to compressing a gzipped file, it leads to small reductions if any, or even increases.

dougwilson:
Yea, reading more I agree that's what it's talking about: the internal compression inside the format, not the file over a transport like HTTP, which is what the "compressible" flag is describing.

dougwilson:
From the linked page:

> WOFF has built-in compression. Ensure that your WOFF compressor is using optimal compression settings.

Things that are internally compressed don't compress after that.

PR #116, opened 2018-03-15
Update to reflect `application/ecmascript`

snuggs:
Thank you for your time! 🙏 ❤️

## Existing `mime-db` IANA Definition (without extension)

 - [x] https://github.com/jshttp/mime-db/blob/master/src/iana-types.json#L464-L468

## References
- [x] Source IANA Registration (with extension) - http://www.iana.org/assignments/media-types/application/ecmascript
- [x] ECMASCript Media Type Specification (with extension) (rfc4329 Section 8.2) - https://tools.ietf.org/html/rfc4329#section-8.2
- [x] WHATWG valid Javascript Mime Types - https://html.spec.whatwg.org/#javascript-mime-type
- [x] WHATWG MIME Sniffing - https://mimesniff.spec.whatwg.org/#javascript-mime-type

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -24,7 +24,12 @@
     ]
   },
   "application/ecmascript": {
-    "compressible": true
+    "compressible": true,
+    "extensions": ["es"],
+    "sources": [
+      "http://tools.ietf.org/rfc/rfc4329.txt",
+      "http://www.iana.org/assignments/media-types/application/ecmascript"
+    ]
   },
   "application/edi-x12": {
     "compressible": false


snuggs:
Thanks @dougwilson! How do I (vicariously) merge?

dougwilson:
Hi @snuggs sorry I didn't see your reply right away. I merged your PR 👍 I'm not 100% sure what the question you're asking is, though.

snuggs:
oh was merely how do i merge Without having merge rights (vicarious). Seems strange. Says this PR was open and closed instead of merged. Also I don't know how this works but usually order is of importance in arrays. I was informed that if this is the case `es` needs to be first. Actually `ecma` is no longer used. Shall I make the update?

Apologies for the inconvenience.

dougwilson:
> oh was merely how do i merge Without having merge rights (vicarious).

Gotcha. As far as I know, that's not possible on GitHub.

> Says this PR was open and closed instead of merged. Am asking as this removes PR merge from my contributions. I like to get 'em where I can. :-)

Not sure. I don't know how to fix, but if you know, please let me know :) ! and I can definitely fix it to appear as merged.

> Also I don't know how this works but usually order is of importance in arrays. I was informed that if this is the case es needs to be first. Actually ecma is no longer used based on the same spec.

Anything in custom is always the lowest precidence. Because the extension "ecma" appears in the Apache project (http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types) we're going to prefer that more. We'll automatically pull in any changes made in the Apache project if you think it's worth fixing there.

> We also would need to add .mjs there as I saw a previous PR come in related. Shall I make the update?

I believe `.mjs` is already added, right? There was a PR I merged about that many months ago.

> And if so what do we prefer to add out of what was mentioned (if any)?

I'm lost on what this question is related to, sorry!

dougwilson:
> Also I don't know how this works but usually order is of importance in arrays.

Also, additional thought here is why is it important? Is there something missing from our documentation regarding the meaning of the `.extensions` array? Here is what our docs say currently:

> .extensions[] - known extensions associated with this mime type.

It doesn't seem to imply any real meaning to the extensions, so not sure if it actually matters for something.

snuggs:
> I believe .mjs is already added, right? There was a PR I merged about that many months ago.

That was for `text/javascript` . There are a bunch others actually.

dougwilson:
Gotcha. Yea, feel free and provide a link to the specifications that show the extension belongs to the type :+1: just like you did here.

PR #121, opened 2018-04-25
Update "application/vnd.citationstyles.style+xml"

rmzelle:
Add .csl extension, add URL to IANA registration page.

I wasn't sure about the "compressible" setting (these files always consist of plain XML, so I guess it makes sense to compress them?). Is `"compressible": false` only for files that wouldn't benefit much from additional compression, like zip files?

This is for Citation Style Language (CSL) citation styles that have an .csl extension. I registered the "application/vnd.citationstyles.style+xml" MIME type myself at IANA. See https://www.iana.org/assignments/media-types/application/vnd.citationstyles.style+xml.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -170,6 +170,7 @@
     ]
   },
   "application/vnd.citationstyles.style+xml": {
+    "compressible": true,
     "extensions": ["csl"],
     "sources": [
       "https://www.iana.org/assignments/media-types/application/vnd.citationstyles.style+xml"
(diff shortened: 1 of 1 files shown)


Martii:
> these files always consist of plain XML, so I guess it makes sense to compress them?

Yes.

> Is "compressible": false only for files that wouldn't benefit much from additional compression

Correct.

Technically most everything is compressible but can make it a waste of decompress time if it just stores it or minute savings. Only other situation is where a format loses integrity but that's extremely really rare.

rmzelle:
@Martii, thanks for confirming. Added `"compressible": true` for this XML-based file type.

Issue #114, opened 2017-12-29
Compressible definition for more mimetypes


danny0838:
Follow the same logic in #110. I think we can add to custom-suffix.json that "+xml" as compressible and "+zip" as not compressible.

dougwilson:
SGTM. For anyone who wants to make a PR on this, please split between two PRs :+1:

rmzelle:
> I think we can add to custom-suffix.json that "+xml" as compressible and "+zip" as not compressible.

@danny0838, @dougwilson, I was wondering if this should be applied just to the existing contents of custom-suffix.json, or if new entries should also be added to custom-suffix.json to cover any aggregated data that ends up in db.json from other sources? If the former, the only chance to custom-suffix.json that's currently necessary would be to add `"compressible": true` to the definition of "application/dash+xml" (see https://github.com/jshttp/mime-db/pull/122/commits/638857e1817677818d33846bb11fc6e0504bfd0d).

dougwilson:
> I was wondering if this should be applied just to the existing contents of custom-suffix.json

I'm not really following. The only thing in there is just the +json.

> or if new entries should also be added to custom-suffix.json to cover any aggregated data that ends up in db.json from other sources?

Isn't that what the request for this issue is?

> If the former, the only change to custom-suffix.json that's currently necessary would be to add "compressible": true to the definition of "application/dash+xml"

I'm not sure what this means. There is only one thing in custom-suffix.json and it's not application/dash+xml

rmzelle:
Oh, sorry for the confusion. Where I wrote "custom-suffix.json" above I meant "custom-types.json".

rmzelle:
Ah, I also realize I misunderstood @danny0838's proposal above. Just ignore me for now. Proper PR coming up.

PR #122, opened 2018-04-27
Update custom-types.json

rmzelle:
* Standardize formatting
* Make XML formats compressible

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -550,13 +550,13 @@
   },
   "image/jp2": {
     "compressible": false,
-    "extensions": ["jp2","jpg2"],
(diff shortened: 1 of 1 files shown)


rmzelle:
I was working on https://github.com/jshttp/mime-db/issues/114, but I had some trouble to get the parsed JSON back into your current format. Do you use a reformatter like https://jsonlint.com/? I couldn't find one that produced your exact format.

For now I just fixed some formatting inconsistencies by hand. I only located one +xml type in custom-types.json that didn't have `"compressible"=true` (added in second commit), and couldn't find any +zip types in custom-types.json.

PR #123, opened 2018-04-29
Mark all +xml types as compressible

rmzelle:
Partially addresses #114.

### DIFF
--- a/src/custom-suffix.json
+++ b/src/custom-suffix.json
@@ -2,5 +2,9 @@
   "+json": {
     "compressible": true,
     "notes": "+json means it is JSON, which is compressible."
+  },
+  "+xml": {
+    "compressible": true,
+    "notes": "+xml means it is XML, which is compressible."
   }
 }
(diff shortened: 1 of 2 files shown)


rmzelle:
Done. Do you still wish to keep entries that now only have a "sources" property? E.g.:

```
  "application/atom+xml": {
    "sources": [
      "http://en.wikipedia.org/wiki/Atom_(standard)"
    ]
},
```

I guess it's still useful in providing a documentation link to the MIME type in question.

dougwilson:
> Do you still wish to keep entries that now only have a "sources" property?

You can delete them. There is no purpose to them at that point. Without any data in the object, the source doesn't reference where the data came from. The purpose of the source is to reference where the data in the object came from.

dougwilson:
For your example, we already have a list of source: https://github.com/jshttp/mime-db/blob/master/src/iana-types.json#L118-L124 all of which are more authoritative than Wikipedia.

rmzelle:
Done. I also found two +json types that still had `"compressible": true`, so I deleted those as well.

dougwilson:
Please move the jsom changes to a separate PR and I can merge this as just the XML changes.

rmzelle:
Cleaned up.

PR #124, opened 2018-04-29
Clean up +json entries in custom-types.json

rmzelle:
* Remove redundant "compressible" property for +json types (since this is already defined for all +json types in custom-suffix.json)
* (removed from PR) Remove "application/fido.trusted-apps+json" entry since it only has the "sources" property

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -31,6 +31,11 @@
   "application/edifact": {
     "compressible": false
   },
+  "application/fido.trusted-apps+json": {
+    "sources": [
+      "https://fidoalliance.org/specs/fido-u2f-v1.0-ps-20141009/fido-appid-and-facets-ps-20141009.html"
+    ]
+  },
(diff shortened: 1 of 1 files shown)


rmzelle:
> 1) mime-db should not contain types not in src/:
     AssertionError: type "application/fido.trusted-apps+json" should be in src/

My guess is that "application/fido.trusted-apps+json" is only defined in custom-types.json. I deleted it because it only has a "sources" property in custom-types.json (per your suggestion at https://github.com/jshttp/mime-db/pull/123#issuecomment-385218749). I assume we should keep it, then?

See https://github.com/jshttp/mime-db/blob/96ceb40103a51b99075fd0a0d57e3a219a6dc0c3/db.json#L424 for the corresponding entry in db.json.

rmzelle:
I see it was added in https://github.com/jshttp/mime-db/pull/85. I'll restore the entry.

PR #125, opened 2018-04-29
 Mark all ZIP-archive types as not compressible

rmzelle:
Addresses the second half of #114. Sibling PR of #123.


rmzelle:
There are no +zip MIME types in custom-types.json, so this should be good to go.

PR #129, opened 2018-06-06
Adding support for HEIC/HEIF Image Formats

### LINKED ISSUE #93: No support for HEIC / HEIF
with iOS 11 this is going to be a big issue..

aspencer:
Hopefully this is in order. Should resolve #93. Let me know if you need any changes!

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -514,14 +514,16 @@
     "extensions": ["heic"],
     "notes": "High Efficiency Image Format (HEIF) Image",
     "sources": [
-      "https://nokiatech.github.io/heif/technical.html"
+      "https://nokiatech.github.io/heif/technical.html",
+      "https://github.com/nokiatech/heif/wiki"
     ]
   },
   "image/heif": {
     "extensions": ["heif"],
     "notes": "High Efficiency Image Format (HEIF) Image",
     "sources": [
-      "https://nokiatech.github.io/heif/technical.html"
+      "https://nokiatech.github.io/heif/technical.html",
+      "https://github.com/nokiatech/heif/wiki"
     ]
   },
   "image/jpeg": {
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -510,6 +510,20 @@
   "image/gif": {
     "compressible": false
   },
+  "image/heic": {
+    "extensions": ["heic"],
+    "notes": "High Efficiency Image Format (HEIF) Image",
+    "sources": [
+      "https://nokiatech.github.io/heif/technical.html"
+    ]
+  },
+  "image/heif": {
+    "extensions": ["heif"],
+    "notes": "High Efficiency Image Format (HEIF) Image",
+    "sources": [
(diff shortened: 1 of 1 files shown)


dougwilson:
Did anyone submit the request to the IANA as mentioned in the issue you linked to?

aspencer:
Hm. I'm sorry, I'm not sure I follow. I don't think there was a reference to the IANA on that issue.

dougwilson:
To get it added, follow steps at https://github.com/jshttp/mime-db#adding-custom-media-types

aspencer:
Ah, I see. No, as far as I can tell it hasn't been added to IANA, and I don't know enough about the format to make a request myself. 

I've just encountered the format "in the wild", so was attempting to add a custom type according to Contributing. Apologies for the misunderstanding. Feel free to close if that's the proper course of action.

dougwilson:
Gotcha. Our goal is just to aggregate from the three sources listed in the README (IANA, Apache, NGINX) and simply present it in a nice format. There isn't the manpower nor desire to administrate a source database itself, though (it was tried for a bit, but no longer). Ideally if this really is used it would be in at least one of those three places (especially the IANA, which is the administrator for this exact type of thing). I hope that makes sense.

aspencer:
Totally does. I'm surprised that this format isn't anywhere, to be honest, as it's the default image format on new iOS devices. Bummer though, going to need to monkeypatch around `mime-types` to get something working - Would have been nice to update at the source! 

Feel free to close. Thanks for your time.

dougwilson:
Yea, I would think so too. Technically this module is _not_ the source, just another middleman like `mime-types`. The source is either IANA, Apache, or NGINX. This module pulls from those sources and simply aggregates them together for further consumption.

aspencer:
Ha! Well, I understand that position, but as far as my code is concerned it might as well be the source. I would make an argument that there's value in a strictly-controlled "community" supplement to the official sources, but that's likely because I'm not doing that work. ;)

dougwilson:
Yes, the IANA is a community source, in fact. I'm just one person maintaining this aggregator module, certainly no where close to the definition of a community and no community resources. The IANA form goes to a mailing list with dozens of active folks.

aspencer:
Great point. Anyways, thanks again for your time.

dougwilson:
From https://tools.ietf.org/html/rfc6838#section-5 :

> The media type registration procedure is not a formal standards
   process, but rather an administrative procedure intended to allow
   community comment and sanity checking without excessive time delay.

Formal registrations require an entire procedure, but simply submitting via the form is not the formal process and just a community review process to say either "yep, this looks right! add to database" or bring up some kind of conflict or other issue :)

dozoisch:
@dougwilson https://www.iana.org/assignments/media-types/media-types.xhtml they seem to have been added in July. Should this PR be re-opened and merged?

dougwilson:
It's been merged. See the commit referenced above your comment.

PR #132, opened 2018-06-10
added support for composer (PHP)

oscarotero:
Hi.
This change allows this package to be installed with php, so PHP projects can download and use the database.
Because php packages use tags for versions, I've added also a `.gitattributes` file to ignore some files on generate the tag file.
Let me know what do you think about this.
Thanks!


dougwilson:
Hi @oscarotero thanks for the PR! I'm not too familiar with PHP stuff these days, but at least for those `.gitattributes` changes, would that affect the files that are being included in the GitHub tarball downloads? In the Node.js community, those need to be "full fat" and include all tests files, etc. for the CITGM (https://github.com/nodejs/citgm) tool to function correctly.

oscarotero:
Ah, got it. Yes, `.gitattributes` affects to how git creates the tarball files for tags.
No problem, I can move that stuff to `composer.json`.

To include a package to PHP, there are three steps:
* create an account in packagist (https://packagist.org/),
* submit this repository (https://packagist.org/packages/submit)
* Configure a service hook in the github repository settings (https://packagist.org/about#how-to-update-packages)

dougwilson:
Thanks! I'm not interested in making an account and managing another package system, honestly. I'm stretched thin as it is. Happy to accept a PR but that's pretty much where I would end, though.

oscarotero:
Ok, I understand.
Thanks anyway.

PR #138, opened 2018-07-31
Add text/event-stream with compressible: false

shesek:
Server-sent events should not be buffered by the server and therefore should not be compressed.

Currently, the express middleware `compression` prevents SSE messages from going through properly. This should fix it.

### DIFF
--- a/db.json
+++ b/db.json
@@ -6918,6 +6918,9 @@
   "text/enriched": {
     "source": "iana"
   },
+  "text/event-stream": {
+    "compressible": false
+  },
   "text/fwdred": {
     "source": "iana"
   },


dougwilson:
This is likely the wrong place for this information.  Not buffering on a server is different from compression flag here.

Specifically the compression boolean is defined in the readme:

> compressible - leave out if you don't know, otherwise true/false to indicate whether the data represented by the type is typically compressible.

shesek:
I would say that because `text/event-stream` is meant for real-time streaming data, it is by definition not suitable for compression.

Would you agree that the current out-of-the-box behavior by `express` + `compression` is problematic? If so, where would be a better place to add this fix?

dougwilson:
I don't understand exactly the issue without like providing some kind of reproduction,  but your description doesn't make sense here. Maybe can you paste here what typical content would be under that MIME type? I can see if it can be compressed or not. Usually the things that are not compressible are things that have already been compressed (like png or jpeg as examples).

If there is a lib you're having an issue with, opening an issue with that lib directly may get the solution or answers you're looking for.

If this content is text (which is compressible), then it makes no sense to say it's not compressible based on the definition the field represents in in our db.

The compression lib seems to include a example right in the readme about how to use the mime type you have here. https://www.npmjs.com/package/compression#server-sent-events Does that example not work? I would open an issue on that module especially since that module is directly talking about the mime you're talking about here.

shesek:
> Maybe can you paste here what typical content would be under that MIME type? 

`text/event-stream` consists of textual contents that usually is compressible, but my thinking was that since `text/event-stream` is expected to be streamed live, it would never make sense to compress it. But: (a) I agree with you that `mime-db`'s `compressible` field is not suitable here, where the reason its not compressible is not due to the contents type itself but rather due to the way its transported and (b) I'm now not entirely sure that streaming data necessarily contradicts compression.

In any case, the solution you linked to in `compression`'s README makes more sense than what I was trying to do. I'm not sure how I missed that! This resolves my issue. Thank you and apologies for the misguided PR :)

PR #146, opened 2018-11-08
Added .acl file support

boulderwebdev:
I noticed there is no support for .acl files, which are permissions files used by the solid platform. For more information, check out 
* https://www.w3.org/wiki/WebAccessControl
* https://github.com/solid/web-access-control-spec

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -666,6 +666,15 @@
   "multipart/signed": {
     "compressible": false
   },
+  "text/acl": {
+    "compressible": true,
+    "extensions": ["acl"],
+    "sources": [
+      "https://www.w3.org/wiki/WebAccessControl",
+      "https://github.com/solid/web-access-control-spec"
+    ],
+    "notes": "Decentralized file permissions"
+  },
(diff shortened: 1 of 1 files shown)


boulderwebdev:
Unfortunately, there is no standard mime-type for .acl files since the standard is relatively new. I am adding this to fix a bug in https://github.com/solid/node-solid-server.

dougwilson:
Hi @boulderwebdev that's a bummer. This module is just an aggregation service, and not the place to introduce a new MIME type to the general web. You'll want to register a MIME type with the IANA first, at which point this module would actually just automatically pull it in.

dougwilson:
You can find more about this process at https://github.com/jshttp/mime-db#adding-custom-media-types

> The best way to get new media types included in this library is to register
them with the IANA. The community registration procedure is outlined in
[RFC 6838 section 5](http://tools.ietf.org/html/rfc6838#section-5). Types
registered with the IANA are automatically pulled into this library.

boulderwebdev:
Awesome! Thanks for the help

PR #147, opened 2018-11-29
Add opus extension

mtAlves:


Martii:
Page 30 is where this is mentioned.  :+1: from me.

Ref:
* https://tools.ietf.org/html/rfc7845

dougwilson:
Agreed @Martii . One strange this is that this has a parameter, which is strange. We have never had one of those before and need to understand our consumers to add this in without a major version bump / change to our database format.

One concern is that there is no defined format to relay parameters in a media type. Of course other specs like the HTTP Content-Type defines how to format parameters with a media type, but the base spec I usually refer to for the format, RFC 6838 (the one linked in the README of this module) gives no format specification for how to represent them together.

I think if we need to represent parameters in the media type, we need to add something new to the database format to hold this information and leave it up to the higher levels (like HTTP, for example) that provide specifications for how to join parameters and media types together.

PR #149, opened 2019-01-04
Support for application/x-zip-compressed and application/x-zip

hdwatts:
When a Windows user converts a folder to a zip file, the filetype becomes: `x-zip-compressed` as shown below (using the `mime-types` npm package, which utilizes this package):
image

There are a few sources, however I am having trouble finding one that isn't an aggregated list:

Closest I can find is SitePoint:
https://www.sitepoint.com/mime-types-complete-list/
http://filext.com/file-extension/ZIP

My company will be using my forked branch in the meantime, as we have windows users uploading zip files pretty regularly.  Let me know if there is anything else I can do.

### DIFF
--- a/db.json
+++ b/db.json
@@ -5583,6 +5583,12 @@
     "source": "apache",
     "extensions": ["xz"]
   },
+  "application/x-zip": {
+    "extensions": ["zip"]
+  },
+  "application/x-zip-compressed": {
+    "extensions": ["zip"]
+  },
   "application/x-zmachine": {
     "source": "apache",
     "extensions": ["z1","z2","z3","z4","z5","z6","z7","z8"]
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -434,13 +434,13 @@
   "application/x-zip": {
     "extensions": ["zip"],
     "sources": [
-      "http://filext.com/file-extension/ZIP"
+      "https://www.oreilly.com/library/view/web-design-in/0596009879/ch04s05.html"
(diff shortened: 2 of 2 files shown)


hdwatts:
I really don't think there is any available primary source documentation, it is a Microsoft mime-type for .zip files created within windows.

image
`HKEY_CLASSES_ROOT\.zip` in regedit, Windows 10 version 1803

This is the default registry value, and appears to have been for [over 9 years](https://bugzilla.mozilla.org/show_bug.cgi?id=540900).  Why IANA hasn't been updated I do not know.

I've added exceptions to handle this situation within my company's codebase, so do what you will with this PR with no primary source.  However, this is a valid mime-type and I worry the buggy situations that my company has experienced will occur for other users of both this library and the multitude of libraries that utilize it.  I'd hope you share in this concern.

dougwilson:
I agree that it's likely been around for a long time. However, the goal of this module is simply to aggregate the three sources in the readme together. We have some built in ones for legacy reasons, but for every person who wants to add one (like you) someone will then come back later saying it should actually be something different.

I do not want to be a mediator here.

Without a primary source it won't be accepted here directly. The best best is to get it added to one of the three sources we aggregate from and so it'll just be added in here automatically.

cjtafoya:
Ridiculous that you would close this PR. Known issue with the library caused by excluding the mimetype generated by the most popular OS on the planet. @hdwatts has a valid case for a known issue...

PR #152, opened 2019-01-20
Updated README from using RawGit to jsDelivr

bodaz:
As RawGit is in sunset phase, we should change the link in README to using a more reliable service like `jsDelivr ` instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,12 @@ npm install mime-db
 ### Database Download
 
 If you're crazy enough to use this in the browser, you can just grab the
-JSON file using [RawGit](https://rawgit.com/). It is recommended to replace
-`master` with [a release tag](https://github.com/jshttp/mime-db/tags) as the
-JSON format may change in the future.
+JSON file using [jsDelivr](https://www.jsdelivr.com/). It is recommended to
+replace `master` with [a release tag](https://github.com/jshttp/mime-db/tags)
+as the JSON format may change in the future.
 
 ```
-https://cdn.rawgit.com/jshttp/mime-db/master/db.json
+https://cdn.jsdelivr.net/gh/jshttp/mime-db@master/db.json
 ```
 
 ## Usage


README.md
```diff
@@ -24,12 +24,11 @@ npm install mime-db
 ### Database Download
 
 If you're crazy enough to use this in the browser, you can just grab the
-JSON file using [RawGit](https://rawgit.com/). It is recommended to replace
-`master` with [a release tag](https://github.com/jshttp/mime-db/tags) as the
-JSON format may change in the future.
+JSON file using [unpkg](https://unpkg.com/mime-db/db.json). It is recommended 
+to keep the version in the URL as the JSON format may change in the future.
```
dougwilson: Can you add instructions or a link to instructions on how to actually put the version in the URL? Based on this sentence and the given example I truthfully have no idea how to put the version in there.

bodaz: Understand. I've updated the instructions now

README.md
```diff
@@ -24,12 +24,18 @@ npm install mime-db
 ### Database Download
 
 If you're crazy enough to use this in the browser, you can just grab the
-JSON file using [RawGit](https://rawgit.com/). It is recommended to replace
-`master` with [a release tag](https://github.com/jshttp/mime-db/tags) as the
-JSON format may change in the future.
+JSON file using [unpkg](https://unpkg.com/mime-db/db.json).
```
dougwilson: Sorry I didn't see this one the first time. I think this should just be a link to the unpkg page for folks to learn more, rather than clicking a link to the json, right?

README.md
```diff
@@ -24,12 +24,18 @@ npm install mime-db
...
 ```
-https://cdn.rawgit.com/jshttp/mime-db/master/db.json
+# use the latest database 
+https://unpkg.com/mime-db/db.json
+```
+
+It is recommended to keep the version in the URL as the JSON format may change in the future.
+
+```
+# always use the v1.37.0 version
```
dougwilson: We use to have an example with a specific version, but people copy it as is and wanted us to keep updating the version in the readme with every release. Instead it was a generic instruction. Can you make it like that with unpkg like it was with rawgit?

bodaz: Sure. How about the latest updates?

dougwilson: Is it not possible to link to the list of versions like it used to? Also not as big of a deal, but it would be nice not to even have the 1 part, as it is too easy to publish a new version and forget this is there.

bodaz: Sure. I found that `jsDelivr` provides release tags in the URL like RawGit does so I've updated the links to `jsDelivr` now, the instructions are very similar to the RawGit one now.

PR #155, opened 2019-02-03
Add extensions for Turtle and N-Triples

pheyvaer:
Hi,

I've added the extensions for `application/n-triples` and `application/n-quads`.


pheyvaer:
My bad! I didn't check out the contribution section in the README. 😅

I read that the IANA types are automatically added, but the extension for `application/n-triples` is not added, while it is defined on IANA [here](https://www.iana.org/assignments/media-types/application/n-triples). Is this then bug in the code?

dougwilson:
So for IANA getting the extensions are complex, as they are just in those documents and not in the main CSV. Just add to custom types json and link to that iana doc as the source :+1:

pheyvaer:
Made requested changes :)

dougwilson:
Thanks! I'm not sure what you did to the `db.json` file, but I'll push up to your branch a revert of all those property changes.

PR #159, opened 2019-04-04
Add extensions .siv and .sieve to application/sieve

MihaiB:
- I copied the `application/sieve` entry from `src/iana-types.json`.
- I changed `http` to `https` for both `sources` (the same documents are served over https too).
- I added the 2 `extensions`.
Up to you which extension should be preferred (first in the array): you know what that means and how it's used:
  - First chronologically? Both sources list them in the order `.siv`, `.sieve` and the RFC says `.sieve` was added later.
  - The most semantic (most clear) extension? To me that's `.sieve`.
  - To resolve collisions? Other words (MIME types) could conceivably be abbreviated to `.siv`, but for each request the webserver has to pick one MIME type.
- `npm run build`
- I added an entry to HISTORY.md as I saw in other recent commits.

https://help.github.com/en/articles/mime-types-on-github-pages says this repository is used by GitHub Pages. I've put a small text file `filter.sieve` on GitHub Pages: it's now served as
```HTTP
content-type: application/octet-stream
```
Did I understand correctly: if my change is accepted, GitHub Pages will serve such files as `application/sieve` soon afterwards?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Add extensions `.siv` and `.sieve` to `application/sieve`
   * Add new upstream MIME types
 
 1.38.0 / 2019-02-04
--- a/db.json
+++ b/db.json
@@ -1333,7 +1333,8 @@
     "extensions": ["shf"]
   },
   "application/sieve": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["siv","sieve"]
   },
   "application/simple-filter+xml": {
     "source": "iana",
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -154,6 +154,13 @@
       "https://www.w3.org/TR/owl-ref/#MIMEType"
     ]
   },
+  "application/sieve": {
+    "extensions": ["siv","sieve"],
+    "sources": [
+      "https://tools.ietf.org/rfc/rfc5228.txt",
+      "https://www.iana.org/assignments/media-types/application/sieve"
+    ]
+  },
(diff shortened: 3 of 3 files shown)


dougwilson:
Awesome, thank you @MihaiB ! Unless you have any idea, I would probably just list them in the order listed in the RFC, which is the order you put in here.

As far as GitHub Pages goes, unfortunately this is an independent project. It sounds like they create the mappings from this project, but the timelines as to when this is merged and when GitHub Pages will be updated I have no idea.

MihaiB:
Sounds good, thank you for the review!

MihaiB:
FYI GitHub Pages now serves static `.sieve` files as
```HTTP
content-type: application/sieve
```
instead of `application/octet-stream`.

I keep a few static files on GitHub pages. Browsers display some files (`.txt`, `.pdf`) directly and offer to download others.
My (perhaps naive) goal was to:
- be specific/semantic (`.sieve` not `.txt` if the file is an example of a sieve script)
- serve files with the appropriate `content-type`
-  have the browser display text formats (like it does for `.txt` files) not download them

GitHub Pages pointed me to this repo to add new mime types, and indeed they picked up this change! However, for `application/sieve` in particular, the browsers (Firefox, Chrome) don't display it by default, but still download it (this might be configurable, but I won't go that far).

This comment is a bit off-topic, I just wanted to report back my use-case / intention / how I got here and that it worked. Thank you again for the speedy review and release of this tiny change!

dougwilson:
Awesome, glad to hear the change got picked up by GitHub pages :) !

PR #161, opened 2019-05-29
application/x-protobuf MIME type as compressible

reyemtm:
Used for vector tiles in web mapping applications.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -349,7 +349,9 @@
   },
   "application/x-protobuf": {
     "compressible": true,
-    "sources": ["https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/protobuf/ProtocolBuffers"]
+    "sources": [
+      "https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/protobuf/ProtocolBuffers"
+    ]
   },
     "application/x-rar-compressed": {
     "compressible": false
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -347,7 +347,11 @@
   "application/x-pkcs12": {
     "compressible": false
   },
-  "application/x-rar-compressed": {
+  "application/x-protobuf": {
+    "compressible": true,
+    "sources": ["https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/protobuf/ProtocolBuffers"]
+  },
+    "application/x-rar-compressed": {
     "compressible": false
   },
   "application/x-sh": {


src/custom-types.json
```diff
@@ -347,7 +347,11 @@
   "application/x-pkcs12": {
     "compressible": false
   },
-  "application/x-rar-compressed": {
+  "application/x-protobuf": {
+    "compressible": true,
+    "sources": ["https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/protobuf/ProtocolBuffers"]
+  },
+    "application/x-rar-compressed": {
```
dougwilson: Indent error here

src/custom-types.json
```diff
@@ -347,7 +347,11 @@
   "application/x-pkcs12": {
     "compressible": false
   },
-  "application/x-rar-compressed": {
+  "application/x-protobuf": {
+    "compressible": true,
+    "sources": ["https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/protobuf/ProtocolBuffers"]
```
dougwilson: Is there a source you can add about the compressibility of the type?

reyemtm:
Hmm I did not find anything on that initially. I'll have to do some digging.

src/custom-types.json
reyemtm: not sure what you mean here - I changed the url to be in a new line if that's what you mean

src/custom-types.json
reyemtm: There is lots of discussion on the [mapbox vector tile spec github issues page](https://github.com/mapbox/vector-tile-spec/issues/27) on compression, but a more authoritative source which clearly states that vector tiles (which are protocol buffers) can be compressed or uncompressed, is the GDAL documentation on [Mapbox Vector Tiles](https://gdal.org/drivers/vector/mvt.html).

src/custom-types.json
dougwilson: It seems like there are four spaces now on this line when there used to be only 2 spaces.

ctaylor4874:
Is there anything I can do to help with this PR? I'm also in need of adding `application/x-protobuf` to the compressible types.

dougwilson:
Hi @ctaylor4874 thanks for the ping here. So the PR has an open formatting issue that was never resolved, and the including link results in a 404. But upon further research, it seems that `application/x-protobuf` is a binary container format. So by itself it is not compressible due to the nature of the way in which it encodes the values. That said, of course perhaps the contents in general may be compressible if there are large, compressible strings inside the protobuf contents, just like how a ZIP file can be compressed if the files inside it are compressible but not actually compressed.

Due to this, the actual MIME type `application/x-protobuf` cannot be marked as either yes definately compressible or no definately not compressible in this module, as the MIME type does not provide enough context for the likely compressibility.

PR #165, opened 2019-07-14
Added "ejs" file extension to type "application/javascript".

AaronNGray:
Added "ejs" file extension to type "application/javascript" which is used as well as "mjs" extension for ES6 Modules.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -75,7 +75,7 @@
   "application/javascript": {
     "charset": "UTF-8",
     "compressible": true,
-    "extensions": ["mjs"],
+    "extensions": ["mjs","ejs"],
     "sources": [
       "https://tools.ietf.org/html/draft-bfarias-javascript-mjs-00"
     ]


src/custom-types.json
```diff
@@ -75,7 +75,7 @@
   "application/javascript": {
     "charset": "UTF-8",
     "compressible": true,
-    "extensions": ["mjs"],
+    "extensions": ["mjs","ejs"],
```
dougwilson: Thank you! From https://github.com/jshttp/mime-db#contributing , please add a link in the `sources` here of a primary source that notes that the `.ejs` file extension is the `application/javacript` mime type (just like there is already a source link here for `.mjs`).

AaronNGray: Not to sure about a definitive source, I just have found it used in several articles relating to Babel et.al. I am going to move to .mjs !

AaronNGray:
I will leave this open for a couple of days if you don't mind. Just in case I come up with something definitive.

dougwilson:
Sure, no problem! I'm happy to add it if there is a primary source for the extension to mime linkage, otherwise I don't have any way to verify this, which typically leads to disputes in the future.

dougwilson:
Any updates after this time, or is it ok to close?

AaronNGray:
I found .ejs is actually used for embedded js templating tool https://ejs.co/.

PR #166, opened 2019-07-20
Add text/pug mime type

wvffle:
Since jade changed its name to pug


dougwilson:
Closing due to no response.

PR #167, opened 2019-08-12
Set charset to UTF-8 for image/svg+xml

connorjclark:
first reported here: https://github.com/jshttp/mime-types/issues/66

### DIFF
--- a/db.json
+++ b/db.json
@@ -6680,6 +6680,7 @@
   },
   "image/svg+xml": {
     "source": "iana",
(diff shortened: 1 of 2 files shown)


connorjclark:
It seems that the encoding for an xml document is usually handled by the content itself. Using charset is supported, but it's not preferred and actually complicates things greatly (charset defined, but the XML document specifies a different encoding, then what?) - so the spec  doesn't specify a  default encoding.

closing.

PR #169, opened 2019-08-29
Enable compression for TTF font format

antoniopangallo:
TTF formats are not compressed by default. I think compression should be enabled so that others compression libraries which rely on this db can enable the font/ttf compression by default.


antoniopangallo:
I made the corrections. Is that ok?

PR #170, opened 2019-09-02
Add the dds (DirectDrawSurface) extension to the list of known MIME types

the-simian:
This format is used in Babylon.js applications to store cubemaps for 3d environments. 
here is a live example: https://www.babylonjs-playground.com/#2FDQT5#12

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,7 +3,6 @@
 
   * Add new upstream MIME types
   * Add `application/toml` with extension `.toml`
-  * Add `image/dds` with extension `dds`
   * Mark `font/ttf` as compressible
 
 1.40.0 / 2019-04-20
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 
   * Add new upstream MIME types
   * Add `application/toml` with extension `.toml`
+  * Add `image/dds` with extension `dds`
   * Mark `font/ttf` as compressible
 
 1.40.0 / 2019-04-20
--- a/db.json
+++ b/db.json
@@ -6927,6 +6927,10 @@
     "source": "apache",
     "extensions": ["xwd"]
   },
+  "image/dds" : {
+    "sources": ["https://docs.microsoft.com/en-us/windows/win32/wic/dds-format-overview"],
+    "extensions: ["dds"]
+  },
   "message/cpim": {
     "source": "iana"
   },
--- a/db.json
+++ b/db.json
@@ -6927,7 +6927,7 @@
     "source": "apache",
     "extensions": ["xwd"]
   },
-  "image/dds" : {
+  "image/vnd.ms-dds" : {
     "sources": ["https://docs.microsoft.com/en-us/windows/win32/wic/dds-format-overview"],
     "extensions: ["dds"]
   },
--- a/db.json
+++ b/db.json
@@ -6927,10 +6927,6 @@
     "source": "apache",
     "extensions": ["xwd"]
   },
-  "image/vnd.ms-dds" : {
-    "sources": ["https://docs.microsoft.com/en-us/windows/win32/wic/dds-format-overview"],
-    "extensions: ["dds"]
-  },
   "message/cpim": {
     "source": "iana"
   },
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -637,6 +637,10 @@
   "image/x-xcf": {
     "compressible": false
   },
+  "image/dds" : {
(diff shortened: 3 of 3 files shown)


HISTORY.md
```diff
@@ -3,6 +3,7 @@
 
   * Add new upstream MIME types
   * Add `application/toml` with extension `.toml`
+  * Add `image/dds` with extension `dds`
```
dougwilson: Since 1.41.1 has already been released, you don't want to list this changed under that version. You can just leave this out.

the-simian:
@dougwilson , let me know if there's anything I need to change about this PR. Thank you!

src/custom-types.json
```diff
@@ -637,6 +637,14 @@
   "image/x-xcf": {
     "compressible": false
   },
+  "image/dds" : {
+    "sources": [
+      "https://docs.microsoft.com/en-us/windows/win32/wic/dds-format-overview",
```
dougwilson: I cannot find any reference to `image/dds` on this page. Perhaps I'm missing it, though. I just see `image/vnd-ms.dds`. Can you help me find it?

src/custom-types.json
```diff
@@ -637,6 +637,14 @@
...
+      "https://en.wikipedia.org/wiki/DirectDraw_Surface"
```
dougwilson: You can remove this, as the sources need to be primary courses only, no Wikipedia links (https://github.com/jshttp/mime-db#contributing).

src/custom-types.json
the-simian: actually I might have been mistaken, apparently the MIME type really is `image/vnd.ms-dds`, and not just image/dds, let me fix that.

the-simian: I also confirmed that on this page here: https://www.aconvert.com/image/jpg-to-dds/

src/custom-types.json
the-simian: Tahnk you I'll fix this immediately

HISTORY.md
the-simian: I will remove, thank you.

the-simian:
@dougwilson thank you for such a quick review! I've made all the requested changes!

db.json
```diff
@@ -6927,6 +6927,10 @@
     "source": "apache",
     "extensions": ["xwd"]
   },
+  "image/vnd.ms-dds" : {
```
dougwilson: Please do not make a PR against this file, per instructions at https://github.com/jshttp/mime-db#contributing

If you do want to include it, it should be the output from `npm run build` (this looks like it was done by hand instead).

the-simian: Ok yes I see, I looked at somone elses PR as an example I will remove

the-simian: @dougwilson thanks for all the help, I'm more familiar now with the project, so any future PRs will be better

dougwilson: No problem at all.

src/custom-types.json
```diff
@@ -637,6 +637,11 @@
   "image/x-xcf": {
     "compressible": false
   },
+  "image/vnd.ms-dds" : {
+    "sources": ["https://docs.microsoft.com/en-us/windows/win32/wic/dds-format-overview"],
+    "extensions: ["dds"],
```
dougwilson: It looks like the tests are failing because there is a missing quote on this line, making the JSON file no longer parsable.

the-simian: I also caught that, and its fixed, thank you.

the-simian:
@dougwilson , i know there's been a bit of back and forth, everything looks ok now, its just the one file and the checks have passed. Thanks for being patient with me

dougwilson:
Hi @the-simian can you please check the check box on the PR to allow me to push to your repo? https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

the-simian:
@dougwilson When I arrived it looked checked but I re-toggled it again:
image

dougwilson:
Very strange, then. I just merged another PR just a little bit ago without any issue. For this PR I get the following issue when I'm trying to push to it:

```
ERROR: Permission to simiancraft/mime-db.git denied to dougwilson.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

That check box is what normally gives me access to push to the specific branch that is associated with the PR.

I can still merge the PR, but it will end up showing in the GitHub UI as the red "Closed" instead of the purple "Merged" if I'm not able to update the PR first before landing is all, so not really that big of a deal if we cannot get this to work for your PR.

PR #172, opened 2019-09-18
Remove compressible from multipart/mixed

ralfhandl:
The media type multipart/mixed is essentially a textual media type, it uses "boundary delimiter lines" terminated by CRLF to separate the parts of a request, see https://tools.ietf.org/html/rfc2046#section-5.1:

> The body must then contain
   one or more body parts, each preceded by a boundary delimiter line,
   and the last one followed by a closing boundary delimiter line.
   After its boundary delimiter line, each body part then consists of a
   header area, a blank line, and a body area. 

This media type is used e.g. in Batch request and response bodies as defined by the OASIS and ISO standard Open Data Protocol (OData), see http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part1-protocol/odata-v4.0-errata03-os-part1-protocol-complete.html#_Toc453752315.

In OData the parts of a multipart/mixed message are typically JSON requests and responses, grouped for performance reasons, or to form "change sets" that are committed or rolled back as a logical unit of work.

Examples from the specification:
```
POST /service/$batch HTTP/1.1
Host: host
OData-Version: 4.0
Content-Type: multipart/mixed; boundary=batch_36522ad7-fc75-4b56-8c71-56071383e77b

Content-Length: ###

 

--batch_36522ad7-fc75-4b56-8c71-56071383e77b
Content-Type: application/http

GET /service/Customers('ALFKI')
Host: host


--batch_36522ad7-fc75-4b56-8c71-56071383e77b
Content-Type: multipart/mixed; boundary=changeset_77162fcd-b8da-41ac-a9f8-9357efbbd

--changeset_77162fcd-b8da-41ac-a9f8-9357efbbd
Content-Type: application/http
Content-ID: 1

POST /service/Customers HTTP/1.1
Host: host
Content-Type: application/json
Content-Length: ###

<JSON representation of a new Customer>

--changeset_77162fcd-b8da-41ac-a9f8-9357efbbd
Content-Type: application/http
Content-ID: 2

PATCH /service/Customers('ALFKI') HTTP/1.1
Host: host
Content-Type: application/json
If-Match: xxxxx
Content-Length: ###

<JSON representation of Customer ALFKI>

--changeset_77162fcd-b8da-41ac-a9f8-9357efbbd--
--batch_36522ad7-fc75-4b56-8c71-56071383e77b
Content-Type: application/http

GET /service/Products HTTP/1.1
Host: host


--batch_36522ad7-fc75-4b56-8c71-56071383e77b--
```

A possible response to the above request could be
```
HTTP/1.1 200 Ok
OData-Version: 4.0
Content-Length: ####
Content-Type: multipart/mixed; boundary=b_243234_25424_ef_892u748

--b_243234_25424_ef_892u748
Content-Type: application/http

HTTP/1.1 200 Ok
Content-Type: application/json
Content-Length: ###

<JSON representation of the Customer entity with EntityKey ALFKI>

--b_243234_25424_ef_892u748
Content-Type: multipart/mixed; boundary=cs_12u7hdkin252452345eknd_383673037

--cs_12u7hdkin252452345eknd_383673037
Content-Type: application/http
Content-ID: 1

HTTP/1.1 201 Created
Content-Type: application/json
Location: http://host/service.svc/Customer('POIUY')
Content-Length: ###

<JSON representation of a new Customer entity>

--cs_12u7hdkin252452345eknd_383673037
Content-Type: application/http
Content-ID: 2

HTTP/1.1 204 No Content
Host: host


--cs_12u7hdkin252452345eknd_383673037--

--b_243234_25424_ef_892u748
Content-Type: application/http

HTTP/1.1 404 Not Found
Content-Type: application/xml
Content-Length: ###

<JSON error message>
--b_243234_25424_ef_892u748--
```

So the majority of the multipart/mixed message in these cases is text: the wrapper text of multipart/mixed itself, plus the JSON or XML text in the bodies of the parts. Occasionally intermingled non-text parts do not pose a problem because these octet streams are preserved during compression/decompression and may benefit from compression.

This has been successfully used in thousands of production environments with servers built on Java, .NET, and ABAP stacks, browser-based clients, and native clients on Java (Android), iOS, Java, .NET, and others. It's only our new Node.js-based server that insists on not compressing multipart/mixed messages due to this flag.

Thanks in advance
Ralf Handl

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@ unreleased
 ==========
 
   * Add new upstream MIME types
+  * Remove compressible from `multipart/mixed`
 
 1.41.0 / 2019-08-30
 ===================
--- a/db.json
+++ b/db.json
@@ -7155,8 +7155,7 @@
     "source": "iana"
   },
   "multipart/mixed": {
-    "source": "iana",
-    "compressible": false
+    "source": "iana"
   },
   "multipart/multilingual": {
     "source": "iana"
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -680,9 +680,6 @@
   "multipart/form-data": {
     "compressible": false
   },
-  "multipart/mixed": {
-    "compressible": false
-  },
   "multipart/related": {
     "compressible": false
   },


ralfhandl:
I've added links to the corresponding section of the original RFC of multipart/mixed, as well as links to the OASIS/ISO standard using multipart/mixed successfully with compression. This standard has been implemented by Microsoft, IBM, Red Hat, SAP, and others. 

If you need more links, please tell me.

Thanks in advance
Ralf

dougwilson:
Thanks, but I still see no links in the JSON file?

ralfhandl:
Sorry, I thought it is sufficient that the link to RFC2046 is here already contained in https://github.com/jshttp/mime-db/blob/master/src/iana-types.json#L8747-L8752.

Should I still add additional links in custom-types.json?

dougwilson:
The links currently in the PR I see above are not enough to change the current data on `multipart/mixed`. For example, the RFC says nothing about it being able to be compressed, and really seems to say that each part could contain _any_ data. The other link to the OASIS/ISO standard is about a specific implementation that uses `multipart/mixed`, not of that type itself. Just like an `image/jpeg` could, with a specific data inside of it maybe be compressible, but I wouldn't think that would make `image/jpeg` generally a compressible type, right?

The JSON in the PR should add a `sources` array to the object (https://github.com/jshttp/mime-db#contributing), and the source links themselves should clearly explain that `multipart/mixed`, as a whole, is compressible.

If instead, you think we should remove the compressible flag altogether, that is acceptable as well, since there is no source for the existing flag anyhow.

dougwilson:
> Sorry, I thought it is sufficient that the link to RFC2046 is here already contained in https://github.com/jshttp/mime-db/blob/master/src/iana-types.json#L8747-L8752.

So unless I'm missing it, RFC2046 does not say anywhere inside it is compressible, rather it seems to indicate that ultimately the compressibility depends on what generic data a specific usage is going to place into each part, right?

ralfhandl:
I'm afraid I'm missing your point:
- Multipart/mixed doesn't say anything about compressibility because you can always compress/decompress, and typically this decreases the body size, and in most cases significantly.
- So it does not hurt to compress a multipart/mixed body.
- And it does hurt if it is always sent uncompressed.

dougwilson:
There are three states to the compressible flag in this module:

1. Definitely compressible
2. Definitely not compressible
3. Maybe compressible

This module currently has it set at (2). Your proposal is (1). But I see no evidence of either here, and think it should really be set to (3).

ralfhandl:
Ok, got it. Changed it to `null`, assuming this expresses state (3).

dougwilson:
Unless I'm missing something, `null` is not a documented value https://github.com/jshttp/mime-db#contributing . If you think the contributing docs can be clarified the list what the three values are more, I'm all ears.

dougwilson:
Basically, just update the PR to remove the entire object in this file, since leaving out the compressible property would just leave an empty object, which doesn't do anything anyway.

ralfhandl:
Done.

ralfhandl:
Hi @dougwilson 

Regarding label "awaiting more info": what info is missing?

Thanks in advance
Ralf

dougwilson:
Hi @ralfhandl , the label means the issue is awaiting more info. Obvious it was at the time the label was added. It is used in order to pull up a list of org-wide issues to check on if more info has been provided and if not close the issue if waiting for over 30 days. I haven't had a chance to come back to your issue after our discussion until now, so didn't have the opportunity to remove the label. If I actively needed more info from you, it would have been clear in our discussion.

ralfhandl:
Hi @dougwilson,

Sorry, I didn't want to harass you. 

Thank you for the fast and helpful interaction!

Ralf

dougwilson:
It's no problem. I landed your change and just need to land the PR that was right before you tomorrow and will publish a new version.

PR #173, opened 2019-09-30
Added new mime time

TomaszFoster:
Added Apples new Reality Composer file extension `model/vnd.reality` for `.reality` file types when using ARKit

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -669,6 +669,14 @@
   "model/mesh": {
     "compressible": false
   },
+  "model/vnd.reality": {
+    "compressible": false,
+    "extensions": ["reality"],
+    "notes": "Apple format for Reality Composer. No specs available anywhere.",
+    "sources" : [
+      "https://devstreaming-cdn.apple.com/videos/wwdc/2019/612umedd7bboc1/612/612_advances_in_ar_quick_look.pdf?dl=1"
+    ]
+  },
   "model/vrml": {
     "compressible": false
   },


src/custom-types.json
```diff
@@ -669,6 +669,14 @@
   "model/mesh": {
     "compressible": false
   },
+  "model/vnd.reality": {
+    "compressible": false,
+    "extensions": ["reality"],
+    "notes": "Apple format for Reality Composer. No specs available anywhere.",
+    "sources" : [
+      "https://devstreaming-cdn.apple.com/videos/wwdc/2019/612umedd7bboc1/612/612_advances_in_ar_quick_look.pdf?dl=1"
```
sampathBlam: @TomaszFoster : Sebastian Grassia from Pixar is the one who has registered the usdz model format in IANA. And .reality is a more recent format(introduced in WWDC 2019). It is likely that guys from Apple will register .reality in IANA in sometime. If not you can try registering in IANA using the instructions mentioned in the docs. But I am afraid this PDF which is a slide show of the wwdc presentation cannot be deemed as the primary source of this mime type though it seems to be the only source with both the extension and mimetype mentioned in it.

wesleytodd (changes requested):
The comment fro @sampathBlam makes me think this needs more work. Either way, the `db.json` file change cannot be included as that file is generated. If you would like to continue with this PR please remove that and address @sampathBlam's concerns int he comment.

PR #174, opened 2019-10-06
Add `application/x-keepass2` for KeePass database files

FliegendeWurst:


src/custom-types.json
```diff
@@ -328,6 +328,12 @@
   "application/x-javascript": {
     "compressible": true
   },
+  "application/x-keepass2": {
+    "extensions": ["kdbx"],
+    "sources": [
+      "https://github.com/keepassx/keepassx"
```
dougwilson: I'm not seeing the mime type listed on this page, but perhaps I am missing it. Can you guide me to where it is specified?

FliegendeWurst: In this file: https://github.com/keepassx/keepassx/blob/master/share/linux/keepassx.xml

dougwilson: Please update the link to the specific page which lists the mime type and the file extension.

dougwilson: Thank you! My plane is about to take off now, so will do a full review later today.

PR #175, opened 2019-10-22
Add entries for audio formats supported by android

sampathBlam:
- Add audio types in custom-types.json for gsm, xmf, rtx, rtttl, ota, imy
- Add video types in custom-types.json for webm
- These formats are supported by android (src: https://developer.android.com/guide/topics/media/media-formats)


src/custom-types.json
```diff
@@ -473,9 +473,26 @@
...
+    "sources": [
+      "http://tools.ietf.org/rfc/rfc4856.txt",
+      "http://www.iana.org/assignments/media-types/audio/GSM"
+    ]
+  },
   "audio/l24": {
     "compressible": false
   },
+  "audio/midi": {
+    "extensions":["xmf", "rtttl", "ota", "imy"]
```
dougwilson: Please make sure to add a `sources` array here with links to the primary sources for the association of these extensions to this mime type.

src/custom-types.json
```diff
@@ -473,9 +473,26 @@
   "audio/basic": {
     "compressible": false
   },
+  "audio/gsm": {
+    "extensions": ["gsm"],
```
dougwilson: Neither of the sources for this entry seem to list a file extension of `.gsm` anywhere. I may be missing it though, as the docs are long. Please point out where the file extension is listed for this or add an additional primary source with the link between the file extension and the mime type.

src/custom-types.json
```diff
@@ -493,6 +510,13 @@
   "audio/ogg": {
     "compressible": false
   },
+  "audio/rtx": {
+    "extensions": ["rtx"],
```
dougwilson: Neither of the sources for this entry seem to list a file extension of `.rtx` anywhere. I may be missing it though, as the docs are long. Please point out where the file extension is listed for this or add an additional primary source with the link between the file extension and the mime type.

src/custom-types.json
```diff
@@ -878,7 +902,8 @@
     "compressible": false
   },
   "video/webm": {
-    "compressible": false
+    "compressible": false,
+    "extensions" : ["webm"]
```
dougwilson: Please make sure to add a `sources` array here with links to the primary sources for the association of these extensions to this mime type.

sampathBlam:
@dougwilson : 
I have updated the PR removing the types that 
- Does not have proper primary source / entry in any registry such as IANA/nginx or Apache
- Does have a registration in IANA but does not have the file extension specified in it.  
This PR now covers only xmf, mxmf and webm extensions.

I have a some questions.
1. What qualifies as a primary source?  
2. What should we do for mime types that are registered with IANA but their file extensions are not mentioned in the Additonal information section of registration spec. eg. [GSM](https://www.iana.org/assignments/media-types/audio/GSM), [RTX](https://www.iana.org/assignments/media-types/audio/rtx) 
3. What should we do for file extensions which do not have registrations/entry in all sources such as IANA/nginx or apache ? Eg. RTTTL and imy (iMelody) are known ringtone formats but they do not seem to have any primary source/spec.

If one/more of my questions are beyond the scope of this PR, and demand a separate Q/A issue, please let me know.

dougwilson:
Thanks @sampathBlam 

> What qualifies as a primary source?

The README links to the Wikipedia article about what "primary source" means (https://en.wikipedia.org/wiki/Primary_source). Basically, it boils down to a primary source is the folks who created the thing in the first place. For example, whoever created the RTX type would then write up what the extension is, being the primary source for the information (vs a secondary source).

> What should we do for mime types that are registered with IANA but their file extensions are not mentioned in the Additonal information section of registration spec. eg. GSM, RTX

Contact the original authors to clarify what, specifically, they want the extension to be. Otherwise, follow the IANA registration procedures (https://github.com/jshttp/mime-db#adding-custom-media-types).

> What should we do for file extensions which do not have registrations/entry in all sources such as IANA/nginx or apache ? Eg. RTTTL and imy (iMelody) are known ringtone formats but they do not seem to have any primary source/spec

This module is not a registration service. It is an aggregator. As such, there needs to be a method to vet any random thing. If there is not a primary source, the method to get them into this module is to get them into one of the three sources this module scrapes from, which have vetting processes and folks in place to support such processes. The IANA is the most well-defined one and it's linked to in the README.

src/custom-types.json
```diff
@@ -878,7 +887,12 @@
     "compressible": false
   },
   "video/webm": {
-    "compressible": false
+    "compressible": false,
+    "extensions" : ["webm"],
```
dougwilson: So this is already in NGINX and Apache, and as such, this is already in our database. Adding it here does nothing. Are you having an issue where you believe adding the extension will resolve some kind of issue, and if so, what is the issue you're having?

src/custom-types.json
```diff
@@ -476,6 +476,15 @@
   "audio/l24": {
     "compressible": false
   },
+  "audio/mobile-xmf": {
+    "extensions":["xmf", "mxmf"],
+    "sources": [
+      "http://tools.ietf.org/rfc/rfc4723.txt",
+      "http://www.iana.org/assignments/media-types/audio/mobile-xmf",
+      "https://filext.com/file-extension/XMF",
```
dougwilson: You don't need to include this, as it's not a primary source. The README also specifically calls out that "links to type aggregating sites and Wikipedia are not acceptable."

src/custom-types.json
```diff
@@ -476,6 +476,15 @@
...
+      "http://www.loc.gov/preservation/digital/formats/fdd/fdd000121.shtml"
```
dougwilson: This seems to be a secondary source, especially when the document lists the file extension "XMF" and states that it is "Suggested by the specification". What specification is it referring to as where the XMF file extension comes from?

sampathBlam:
Got it. Thanks for the clarification!

src/custom-types.json
sampathBlam: Oh. I thought I had issues with webm. Apparently not. I will remove this.

src/custom-types.json
sampathBlam: Got it.

src/custom-types.json
sampathBlam: XMF seems to have been created by the MIDI association. [spec](https://www.midi.org/specifications/item/xmf-specification-all). But I could not find a direct mapping between the file extension and the mime-type in their spec though there are references to audio/mobile-xmf. I can remove this as well.

sampathBlam:
@dougwilson : The PR has been updated and now it holds changes only for mxmf format. This PR has provided a lot of learning. Thanks and cheers. I will see what I can do for the other formats.

sampathBlam:
@dougwilson : Squashed all the previous commits into a single one to facilitate easier review.

PR #178, opened 2019-11-30
add typescript

zhmushan:


src/custom-types.json
```diff
@@ -171,6 +171,12 @@
     ],
     "compressible": true
   },
+  "application/typescript": {
+    "source": ["https://github.com/microsoft/TypeScript"],
```
dougwilson: Please make sure the source shows the source of the mime type and any extensions. I cannot find anywhere on this page a reference to the mime type `application/typescript`, but it's long and I may be missing it. Can you direct me to where on this page the mime type is listed?

zhmushan: `typescript` doesn't seem to register a `mime-type`. I found a [backlog](https://github.com/microsoft/TypeScript/issues/25426) in their repo.

PR #180, opened 2019-12-21
Add application/x-python-code and text/x-python

jdek:
Noticed they weren't here since `.py` files are being downloaded instead of displayed in the browser on GitHub pages.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -136,6 +136,12 @@
   "application/postscript": {
     "compressible": true
   },
+  "application/x-python-code": {
+    "extensions": ["pyc", "pyo"],
+    "sources": [
+      "https://github.com/python/cpython/blob/19a3d873005e5730eeabdc394c961e93f2ec02f0/Lib/mimetypes.py#L457"
+    ]
+  },
   "application/raml+yaml": {
     "compressible": true,
     "extensions": ["raml"],
@@ -856,6 +862,12 @@
     "extensions": ["pde"],
     "sources": [
       "https://en.wikipedia.org/wiki/Processing_(programming_language)"
+    ]
+  },
+  "text/x-python": {
+    "extensions": ["py"],
+    "sources": [
+      "https://github.com/python/cpython/blob/19a3d873005e5730eeabdc394c961e93f2ec02f0/Lib/mimetypes.py#L528"
     ]
   },
   "text/x-sass": {
--- a/db.json
+++ b/db.json
@@ -5653,6 +5653,9 @@
     "source": "apache",
     "extensions": ["p7r"]
   },
+  "application/x-python-code": {
+    "extensions": ["pyc","pyo"]
+  },
   "application/x-rar-compressed": {
     "source": "apache",
     "compressible": false,
@@ -7611,6 +7614,9 @@
   "text/x-processing": {
     "compressible": true,
     "extensions": ["pde"]
+  },
+  "text/x-python": {
+    "extensions": ["py"]
   },
   "text/x-sass": {
     "extensions": ["sass"]


src/custom-types.json
```diff
@@ -136,6 +136,12 @@
   "application/postscript": {
     "compressible": true
   },
+  "application/x-python-code": {
+    "extensions": ["pyc", "pyo"],
+    "sources": [
+      "https://github.com/python/cpython/blob/19a3d873005e5730eeabdc394c961e93f2ec02f0/Lib/mimetypes.py#L457"
```
dougwilson: This file is just an aggregate of mime types that relate to all kinds of things. What we want here is a more definitive source. Perhaps some kind of python docs?

jdek: I think typically the source code prevails when looking at documentation vs source code. This source is the official Python source code, and its mime type list.

If you want I can submit a bug to the python bug tracker referencing this issue to ask for 'official documentation' in some capacity?

dougwilson: > I think typically the source code prevails when looking at documentation vs source code.

Normally I would agree, but the source here is a mime type library and it lists an aggregate of unrelated types. It is effectively source to a lib like ours and not the original source.

If instead your proposal here is that we should use that lib as one of our aggregate sources and include all types from there, that would be a different proposal. Ideally just link where they got it from as our link.

> If you want I can submit a bug to the python bug tracker referencing this issue to ask for 'official documentation' in some capacity?

Sure, that is fine. Another method is just to get this mapping included in the aggregates this module pulls from, which is the most ideal solution.

jdek: > Ideally just link where they got it from as our link.

It looks like they just take theirs from IANA or add `*/x-*` types from consensus.

> Sure, that is fine

See https://bugs.python.org/issue39115

dougwilson: > See https://bugs.python.org/issue39115

I think there may be some confusion here. This is module is not meant to be the database where things are added to as the public listing; that is the IANA registry. This module is meant just to pull from the three sources at the top if the readme and aggregate them into a JSON file, with the custom file as a legacy list we are working to get rid of.

In 2.0 we are removing it and if you want python to continue to be listed, that is why I suggested the best method is to get it added to one of our sources.

We can still accept it directly if there is a source provided, but eventually it will drop if it doesn't get added to one of the sources this module pulls from.

I hope that helps clarify what I meant.

jdek: I see, I have also submitted a patch to Apache: https://bz.apache.org/bugzilla/show_bug.cgi?id=64024.

Hopefully it is OK, otherwise might have to see what needs to be done in order to get it registered as an IANA MIME type.

dougwilson: Yep that is good. I will keep an eye on both and update the db if apache lands or see what is said on that python tracker if they respond. You're also welcome to ping here if you see activity in either place.

jdek: I'm not sure it is eligible for an IANA registered type (not to mention they don't think it is necessary for an official one). Also the python library aggregates already from nginx/apache, so these are additional and 'essential' types (if other types can't be found). I'm not really sure where you'd put extra documentation here anyway.

I think the key is just get this into apache as a similar mime to `text/x-c` etc (but I don't really think it's likely).

dougwilson: That's fine whatever you / the source project think is best. This module is not intended to be a source database of types from around the Internet :) That is too big of a commitment from just myself to make. It was originally built to simply pull from the three sources listed at the top of the README and that was it. The custom types was added to be generous, but it has unfortunately been a large burden, of types getting in there and folks arguing it is wrong / etc. So to combat that the only way new things are added to the custom is that there has to be a link to a primary source for the type. The current link just is not clear enough, being a link to a source file that is just an aggregate list of types from all kinds of things.

jdek: I will keep this open until I get a response from the Apache project.

dougwilson:
I'm going to close this for now, mainly because https://bugs.python.org/issue39115 seems to end with declaring the types as "unofficial" specifically. If Python files do actually get some kind of official type we can definitely add them in here if they do not end up in one our the upstreams we pull from by then 👍

dmalan:
Hi @dougwilson, all,

We ran into this issue too on our CDN, whereby `*.py` was defaulting to `binary/octet-stream` as opposed to, e.g., `text/x-python`. Rather than hardcode support for the latter (albeit unofficial), we were inclined to to use https://salsa.debian.org/debian/mime-support/blob/master/mime.types from https://packages.debian.org/sid/mime-support, if it might be compelling to add Debian's package as an additional source for [mime-db](https://github.com/jshttp/mime-db), alongside Apache et al.?

PR #181, opened 2019-12-23
Added Microsoft AppInstaller Mime-Types

shartte:
Microsoft now allows Windows 10 App packages to be distributed from Websites by using appinstaller files and simply publishing the used packages via HTTP.
Sadly, the installer seems sensitive to the correct Mime-Type being sent by the Webserver, which is why they included custom Mime-Types in their setup guide. This adds those mime-types. Luckily they map 1:1 with the custom extensions used for the package files.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -1,4 +1,31 @@
 {
+  "application/appinstaller": {
+    "compressible": true,
+    "extensions": ["appinstaller"],
+    "notes": "Microsoft Windows AppInstaller XML Manifest",
+    "sources": [
+      "https://docs.microsoft.com/en-us/windows/msix/app-installer/web-install-iis",
+      "https://docs.microsoft.com/en-us/windows/msix/app-installer/web-install-azure"
+    ]
+  },
+  "application/appx": {
+    "compressible": false,
+    "extensions": ["appx"],
+    "notes": "Microsoft Windows AppX Package",
+    "sources": [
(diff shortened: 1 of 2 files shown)


src/custom-types.json
```diff
@@ -1,4 +1,31 @@
 {
+  "application/appinstaller": {
+    "compressible": true,
+    "extensions": ["appinstaller"],
+    "notes": "Microsoft Windows AppInstaller XML Manifest",
+    "sources": [
+      "https://docs.microsoft.com/en-us/windows/msix/app-installer/web-install-iis",
```
dougwilson: These links are ideally supposed to link to the primary source for the given extension and mime. This seems to link to a secondary source just documenting how to set up with IIS. A primary source would typically be the spec that would define what an appinstaller file actually is, for example.

dougwilson: For example this says the type is compressible here. I'm not sure how that was determined from either of the links?

shartte: Hi @dougwilson It's true, but it's the only source I actually found that explains which mime types to use to host these files. Except for Microsoft's documentation I was unable to find an actual "spec" of sorts, since it's proprietary Windows-only tech.
I did find that Microsoft maintains a repo on Github with the official docs for MSIX. I'll open an issue there and crosslink it with this one to get some clarity on whether they intended these MIME types to be official or not.

PR #186, opened 2020-02-25
Associate .cjs extension with application/node

wmhilton:
Script files with `.cjs` extensions are interpreted as CommonJS and (from what I can tell) should have the IANA designated `application/node` MIME type, even though the IANA registry entry predates the `.cjs` extension and does not mention `.cjs`. Someone from the https://github.com/nodejs/modules working group probably knows the deets better than me though.

Hopefully I updated the correct file 😄. Thoughts?

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -65,8 +65,9 @@
   "application/javascript": {
     "charset": "UTF-8",
     "compressible": true,
-    "extensions": ["mjs"],
+    "extensions": ["cjs", "mjs"],
     "sources": [
+      "https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_package_scope_and_file_extensions",
       "https://tools.ietf.org/html/draft-bfarias-javascript-mjs-00"
     ]
   },
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -65,9 +65,8 @@
   "application/javascript": {
     "charset": "UTF-8",
     "compressible": true,
-    "extensions": ["cjs", "mjs"],
+    "extensions": ["mjs"],
     "sources": [
-      "https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_package_scope_and_file_extensions",
       "https://tools.ietf.org/html/draft-bfarias-javascript-mjs-00"
     ]
   },
@@ -92,6 +91,9 @@
   },
   "application/msword": {
     "compressible": false
+  },
+  "application/node": {
+    "extensions": ["cjs"]
   },
   "application/octet-stream": {
     "compressible": false,


jkrems:
Well, technically `.cjs` should explicitly *not* be `application/javascript`. It's meant to be [`application/node`](https://www.iana.org/assignments/media-types/application/node). If you have a "YOLO file" that can be seen as either a valid JavaScript script or as a CommonJS file, I'd say that`.js` is the appropriate extension because it effectively means "ambiguous JavaScript-adjacent content; just... try and see what happens".

* `.js`: Originally a JavaScript script but "rarely" used that way anymore. Over time transformed into mostly non-JavaScript things like CommonJS and Compile-to-CommonJS files using import syntax via babel etc.. Most recently also sometimes used for JavaScript modules. From a browser perspective, `application/javascript` is the best we can do and hope that the user knows what they're doing.
* `.cjs`: A recently added clear indication of "this is CommonJS and nothing else". Not meant to be valid JavaScript. `application/node` may prevent accidental use in "pure" JavaScript runtimes like browsers.
* `.mjs`: A recently added clear indication of "this is a JavaScript module and nothing else". Meant to be valid JavaScript. Unfortunately we have to reuse `application/javascript` because there's no unambiguous content type for modules (that doesn't also include scripts).

If something could be a JavaScript script (e.g. UMD), `.js` is the right extension and `.cjs` isn't really applicable.

src/custom-types.json
```diff
@@ -65,8 +65,9 @@
   "application/javascript": {
     "charset": "UTF-8",
     "compressible": true,
-    "extensions": ["mjs"],
+    "extensions": ["cjs", "mjs"],
     "sources": [
+      "https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_package_scope_and_file_extensions",
```
dougwilson: This is required to be a primary source doe these definitions. I'm not sure if it's primary, but more importantly, I don't see it state anywhere that the file extension. cjs is the mime type application/json .

wmhilton:
> If something could be a JavaScript script (e.g. UMD), .js is the right extension and .cjs isn't really applicable.

Ugh. That's annoying. That mean I have to ship a `.js` file in my package that is _not_ a module, despite the fact that `"type": "module"` for my package. 🤔 Well... nothing another directory and another package.json file can't solve 😄

jkrems:
> Ugh. That's annoying. That mean I have to ship a .js file in my package that is not a module, despite the fact that "type": "module" for my package. 🤔 Well... nothing another directory and another package.json file can't solve 😄

It's also possible to not use UMD instead. You can generate a `.js` file that sets globals etc.. As long as it's only loaded by browsers there's no issue. And you can generate a separate `.cjs` file that is CommonJS.

Alternatively, assuming your tooling has proper module support, you can also remove `type: 'module'` and use `.js` for UMD and other script-like things, just like before. :)

src/custom-types.json
```diff
@@ -93,6 +92,9 @@
   "application/msword": {
     "compressible": false
   },
+  "application/node": {
+    "extensions": ["cjs"]
```
dougwilson: Please add a primary source link for this data.

src/custom-types.json
wmhilton: thanks for the quick review! and yep, looks like I was completely mistaken. sorry!

src/custom-types.json
wmhilton: @jkrems ?

GeoffreyBooth:
> That mean I have to ship a .js file in my package that is not a module, despite the fact that "type": "module" for my package.

If Node never loads the file, it doesn’t matter whether it “behaves” according to how the file extension and `"type"` field dictate it should. It could be a JPEG image saved as `image.js`, and as long as Node never loads it, it doesn’t matter that the file’s contents aren’t JavaScript source code. So if you’re already using a build process, just output the file as `.js` and don’t worry about it; and additionally output a `.cjs` file next to it for Node to use.

src/custom-types.json
GeoffreyBooth: The primary source should be https://www.iana.org/assignments/media-types/application/node, but this needs updating to add `.cjs`.

Node’s definition of `.cjs` is here: https://nodejs.org/api/esm.html#esm_enabling

wmhilton:
@GeoffreyBooth Yup, that's the approach I ended up doing. It works, I'm just saddened I gotta ship 3 versions instead of 2 now. If we want to discuss build processes more, lets do it in the [node/modules thread](https://github.com/nodejs/modules/issues/490) though so the MIME folk don't have to wade through it. 😄

src/custom-types.json
GeoffreyBooth: Update: https://www.iana.org/assignments/media-types/application/node now includes `.cjs`.

dougwilson:
Thanks for the update, @GeoffreyBooth . Looks like the checkbox on this PR to allow me to push up the edit is not enabled:
```
ERROR: Permission to wmhilton-contrib/mime-db.git denied to dougwilson.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

ljharb:
@dougwilson unfortunately this is a bug in github; because @wmhilton made the PR from a fork on `wmhilton-contrib`, and not their main account, the "allow edits" checkbox simply doesn't work, even when checked.

@wmhilton, can you invite @dougwilson to your `wmhilton-contrib/mime-db` fork, so they can update the PR? (i'd also recommend in the future not using this pattern, until github fixes it)

dougwilson:
Ah, makes sense. If I don't hear back from the OP, I can still land the PR with changes, it'll just not show up as "merged" in the GitHub UI, which isn't really a big deal anyhow.

PR #187, opened 2020-03-18
Added JSON-LD

stain:
See W3C specification [application/ld+json](https://www.w3.org/TR/json-ld/)

JSON-LD is the RDF format used by http://schema.org/

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -78,6 +78,14 @@
   "application/json5": {
     "extensions": ["json5"]
   },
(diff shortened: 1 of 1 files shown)


dougwilson:
Thank you, but this is already in our database.

jonchurch:
See these lines for reference @stain https://github.com/jshttp/mime-db/blob/e60b4ac7d5ea76d3271a00ea184d7b979b6f878b/db.json#L685-L688

PR #192, opened 2020-04-27
Patched keynote extension

KarimNahas:
The file extension for keynote is incorrect. 
It is marked as `.keynote` instead of `.key` .

### DIFF
--- a/db.json
+++ b/db.json
@@ -2004,7 +2004,7 @@
   },
   "application/vnd.apple.keynote": {
     "source": "iana",
-    "extensions": ["keynote"]
+    "extensions": ["key"]
   },
   "application/vnd.apple.mpegurl": {
     "source": "iana",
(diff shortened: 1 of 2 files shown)


dougwilson:
This is the file extension registered with the IANA: http://www.iana.org/assignments/media-types/application/vnd.apple.keynote

If you think it is wrong, you will need to get the IANA record updated, as this module just sources from there.

KarimNahas:
@dougwilson I've contacted the IANA and they’ve updated their record for `.keynote`. 
Should I reopen the pull request?

dougwilson:
Nice! If it is updated in the IANA then our script will pull it automatically. I'm still seeing it as .key though. Can you link to the updated entry so I can see what is going on?

dougwilson:
Sorry, i was reading backwards... .key is the new one. I will run the update.

PR #196, opened 2020-06-02
Added reStructuredText mime type

vibhash-singh:
Added support to handle reStructuredText files. They have mime-type "text/x-rst" and extension ".rst" .

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -817,8 +817,7 @@
     "compressible": true,
     "extensions": ["rst"],
     "sources": [
-      "https://docutils.sourceforge.io/rst.html",
-      "https://en.wikipedia.org/wiki/ReStructuredText"
+      "https://docutils.sourceforge.io/rst.html"
     ]
   },
   "text/x-sass": {
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -813,6 +813,14 @@
       "https://en.wikipedia.org/wiki/Processing_(programming_language)"
     ]
   },
+  "text/x-rst": {
+    "compressible": true,
+    "extensions": ["rst"],
+    "sources": [
+      "https://docutils.sourceforge.io/rst.html",
+      "https://en.wikipedia.org/wiki/ReStructuredText"
+    ]
+  },
   "text/x-sass": {
     "extensions": ["sass"]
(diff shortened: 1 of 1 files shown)


src/custom-types.json
```diff
@@ -813,6 +813,14 @@
       "https://en.wikipedia.org/wiki/Processing_(programming_language)"
     ]
   },
+  "text/x-rst": {
+    "compressible": true,
+    "extensions": ["rst"],
+    "sources": [
+      "https://docutils.sourceforge.io/rst.html",
+      "https://en.wikipedia.org/wiki/ReStructuredText"
```
dougwilson, line 821: You can remove this, as links to Wikipedia re not accepted (see instructions at https://github.com/jshttp/mime-db/blob/master/README.md#contributing)

src/custom-types.json
```diff
@@ -813,6 +813,14 @@
       "https://en.wikipedia.org/wiki/Processing_(programming_language)"
     ]
   },
+  "text/x-rst": {
+    "compressible": true,
+    "extensions": ["rst"],
+    "sources": [
+      "https://docutils.sourceforge.io/rst.html",
```
dougwilson, line 820: I cannot find the mime type text/x-rst anyone on this page. I may be over looking it, so if you can point out where it is, please do so as a response to this comment.

vibhash-singh: I am not sure why they have not given on the above page but this information is available on Wikipedia page. This file format is more common in the python community. It is mentioned here also https://packaging.python.org/guides/making-a-pypi-friendly-readme/.

I will try to search for other sources.

src/custom-types.json
```diff
@@ -813,6 +813,13 @@
       "https://en.wikipedia.org/wiki/Processing_(programming_language)"
     ]
   },
+  "text/x-rst": {
+    "compressible": true,
+    "extensions": ["rst"],
+    "sources": [
+      "https://docutils.sourceforge.io/rst.html"
```
dougwilson, line 820: So if this doesn't mention the mime type, it would not be a source for the mime type definition. If you want to submit to this module (as opposed to the IANA), we need a first-party source for the mime type.

PR #200, opened 2020-06-26
Reorder the markdown extensions

stof:
md is the most common extension for markdown. And it is also the first listed one in the IANA assignment.
Listing it first helps with tools wanting to choose an extension for a given mime type.

See https://www.iana.org/assignments/media-types/text/markdown for the reference (scroll to the `Additional information` section)

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -721,7 +721,7 @@
   },
   "text/markdown": {
     "compressible": true,
-    "extensions": ["markdown","md"],
+    "extensions": ["md","markdown"],
     "sources": [
       "https://tools.ietf.org/html/rfc7763"
     ]


stof:
see https://github.com/symfony/symfony/pull/37422#discussion_r446060037 for a discussion that triggered this (we extract the default extension for each mime type)

dougwilson:
Thank you. But we do not actually make any promise towards the meaning of the order of the extensions array (see the readme). If the idea is to actually order these in some kind of specific order, we need to first (1) discuss this (2) figure out where to source this information from and (3) update our database.

I will land your PR, but only because it just puts it in the same order of the RFC. I would like to not that, ultimately, you will be relying on behavior that we will just break due to it not being how this database works if the database definition is never actually updated in order to reflect this definition of the extension array order you seem to want.

In addition, you need to provide objective data that supports the claim to which ones are the "most common" and provide a specific definition to what "common" means in this case.

PR #201, opened 2020-06-26
Sort audio/mpeg file extensions to the most common one

YaFou:
The `audio/mpeg` mime is used mostly with the `mp3` extension. This pull request just sorts these file extensions to the most common one. It also adds a new one: `mp1` defined in [the RFC 3003 (page 3)](https://tools.ietf.org/html/rfc3003).
*[Related comment](https://github.com/symfony/symfony/pull/37422#discussion_r446059439)*

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -453,7 +453,7 @@
   },
   "audio/mp3": {
     "compressible": false,
-    "extensions": ["mp3"],
+    "extensions": ["mp3","mpga","mp2","mp2a","m2a","m3a"],
     "notes": "Chromium sends this mimetype instead of audio/mpeg for mp3 files",
     "sources": [
       "https://bugs.chromium.org/p/chromium/issues/detail?id=227004"


dougwilson:
> This pull request just sorts these file extensions to the most common one.

Thank you. But we do not actually make any promise towards the meaning of the order of the extensions array (see the readme). If the idea is to actually order these in some kind of specific order, we need to first (1) discuss this (2) figure out where to source this information from and (3) update our database. Accepting your change would only do you a disservice if we did not do this first, as you will be relying on behavior that we will just break due to it not being how this database works.

In addition, you need to provide objective data that supports the claim to which ones are the "most common" and provide a specific definition to what "common" means in this case.

> It also adds a new one: mp1

Please split into a separate PR.

YaFou:
Thank you for your quick response.

---

> > It also adds a new one: mp1
> 
> Please split into a separate PR.

I will do that.

---

About the sorting, I have already some documents like [the RFC 3003](https://tools.ietf.org/html/rfc3003) which defines the three extensions for `audio/mpeg` (`.mp1`, `.mp2` and `.mp3`). However, [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) defines that the extension is `.mp3`, and [Nginx](http://hg.nginx.org/nginx/raw-file/default/conf/mime.types) also.
What I define as "common" is that the extension which is the most used for this mime around the Internet. `.mp2` is only used now used for broadcasting (see [*MPEG-1 Audio Layer II*](https://en.wikipedia.org/wiki/MPEG-1_Audio_Layer_II) or [*MPEG-2*](https://en.wikipedia.org/wiki/MPEG-2)). Finally, the `.mp1` extension is obsolete since `.mp2` and `.mp3` was released (see [here](https://en.wikipedia.org/wiki/MPEG-1_Audio_Layer_I)).
I think it would be better if the `.mp3` was the first extension for `audio/mpeg` before `mpga`.  

Of course, I'm open to suggestions and discussions and I will be glad to see them.

dougwilson:
> What I define as "common" is that the extension which is the most used for this mime around the Internet. 

Gotcha. So given that definition, please share how you objectively gathered the data for this type, so we can use the same methodology in order to update all other types in the database for this new definition and so we have a method to keep it updated on an ongoing basis (as I believe you definition means that the ordering will change over time as the contents of the Internet changes).

dougwilson:
Closing due to no futher response.

PR #203, opened 2020-07-14
AVIF added

bolknote:


src/custom-types.json
```diff
@@ -515,6 +515,15 @@
       "http://www.phpied.com/gzip-your-font-face-files/"
     ]
   },
+  "image/avif": {
+    "compressible": false,
+    "extensions": ["avif"],
+    "sources": [
+      "https://aomediacodec.github.io/av1-avif/",
+      "https://en.wikipedia.org/wiki/AV1#AV1_Image_File_Format_(AVIF)"
```
dougwilson, line 523: You can leave off this since we don't take Wikipedia sources. The first source is perfect 👍

PR #204, opened 2020-07-19
Add mime to extension mapping audio/flac -> flac

sgpinkus:
DB already has `audio/x-flac` => `flac` mime to extension mapping. This adds `audio/flac` => `flac`. `audio/flac` does not appear to be in IANA, however it is in Apache servers default mime types and listed as the MIME type for FLAC in numerous places. Sources:

  - `audio/flac` is listed in Apache 2 serve mime.types file -- http://apache.mirror.amaze.com.au//httpd/httpd-2.4.43.tar.gz
  - `audio/flac` is listed in the `mime.types` file provided by Debian [mime-support](https://packages.debian.org/buster/mime-support) package. This is actually the default TypesConfig file used by Apache 2 server, and other softwares on Debian.
  - [Mozdev](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Containers#FLAC) states the two MIME types for FLAC are `audio/flac` and `audio/x-flac` (non standard).
  - [Wikipedia](https://en.wikipedia.org/wiki/FLAC) lists `audio/flac` and `.flac` as the extension. Does not mention `audio/x-flac`.

More testing in Debian Linux (these utils probably just use /etc/mime.types DB):

```
> lsb_release -d
Description:	Debian GNU/Linux 10 (buster)
> file --mime-type  sample.flac
sample.flac: audio/flac
> php -v
PHP 7.3.19-1~deb10u1 (cli) (built: Jul  5 2020 06:46:45) ( NTS )
> php -r 'print(mime_content_type("sample.flac") . "\n");'
audio/flac
```

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -455,6 +455,12 @@
   "audio/basic": {
     "compressible": false
   },
+  "audio/flac": {
+    "sources": [
+      "https://wiki.xiph.org/index.php?title=MIME_Types_and_File_Extensions"
+    ],
+    "extensions": ["flac"]
+  },
   "audio/l24": {
     "compressible": false
   },


sgpinkus:
@dougwilson Sorry. Pushed then started reading .. I've redone and force updated this PR.

src/custom-types.json
```diff
@@ -455,6 +455,12 @@
   "audio/basic": {
     "compressible": false
   },
+  "audio/flac": {
+    "sources": [
+      "https://packages.debian.org/buster/mime-support"
```
dougwilson, line 460: This needs to be a primary source; links to type aggregating sites and Wikipedia are not acceptable.

dougwilson:
Thank you @sgpinkus . There is an issue regarding your source I noted above. Once resolved, we can get it landed. Thank you!

src/custom-types.json
sgpinkus: Debian package is not a primary source? This is *the* MIME database for Debian distributions, including Ubuntu.

dougwilson: Sure, I understand that, but they are not the authority on the FLAC mime type, are they? The authority on the type would be the primary source. I assume they sourced this mapping from somewhere, so that would be what to like here; they are by nature (same at this module) not a primary source of this mapping data.

dougwilson: Xiph.Org Foundation seems to be the owners of the FLAC format/specs and it looks like this is the primary source you are looking to link to: https://wiki.xiph.org/index.php?title=MIME_Types_and_File_Extensions

sgpinkus:
There are probably millions of Apache, and other HTTP servers running on Linux using that file. But OK, fair enough, I try to find out exactly how Debian builds their `mime.types` file. Maybe you can add it as a primary source even?

dougwilson:
>  I try to find out exactly how Debian builds their mime.types file

I found the primary source for the mapping and already commented with it above for you.

> Maybe you can add it as a primary source even?

Add what as a primary source?

src/custom-types.json
dougwilson: @sgpinkus I think you may have resolved this on accident, as the link doesn't seem to be replaced still.

sgpinkus:
> I found the primary source for the mapping and already commented with it above for you.

You did? I can't see.

> Add what as a primary source?

This: http://deb.debian.org/debian/pool/main/m/mime-support/mime-support_3.62.tar.gz

dougwilson:
> > I found the primary source for the mapping and already commented with it above for you.

> You did? I can't see.

I think because you marked the conversation as resolved. I unresolved it for you.

> > Add what as a primary source?

> This: http://deb.debian.org/debian/pool/main/m/mime-support/mime-support_3.62.tar.gz

I think there may be confusion on what a "primary source" means. I linked to the Wikipedia description of "primary source" in the README and above: https://en.wikipedia.org/wiki/Primary_source . By definition, that will always be at best a secondary source of information for type data, as the definers of file formats and their specifications are by definition the only primary source for that information.

sgpinkus:
I meant like along side, IANA, Apache, NGINX.

dougwilson:
Gotcha, so not as a primary source, but just somewhere we source from for this database. We could do that, but need to get in touch with whoever manages that list to (1) understand how they source their data and (2) know we're allowed to consume and republish it in our list.

sgpinkus:
@dougwilson this is as close as I can get to a primary source -- https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in#L4158. Is that OK? This is actually the source used in the Debian mime-support package (upstream VCS [here](https://salsa.debian.org/debian/mime-support)).  Nothing in the FLAC spec about its MIME type, only that its extension is [.flac](https://xiph.org/flac/documentation_tools_flac.html).

dougwilson:
Hi @sgpinkus just use the source link I provided in the PR review, unless you think there is something wrong with it (let me know what is wrong with using it).

sgpinkus:
Oooh! Great source! Sorry bad Internet connection ... the drop down wasn't dropping down. Will do. Sec.

sgpinkus:
>  We could do that, but need to get in touch with whoever manages that list to (1) understand how they source their data and (2) know we're allowed to consume and republish it in our list.

FYI @dougwilson I will look into this instead. I think it's the better approach rather than playing whack-a-mole with custom entries. I've emailed the maintainer, awaiting response.

dougwilson:
Alright, that sounds good to me @sgpinkus . I'm not sure what you're timeline for getting the flac change here, but this PR would likely be much faster to get landed than adding an entirely new source :)

sgpinkus:
>  I'm not sure what you're timeline for getting the flac change here, but this PR would likely be much faster to get landed than adding an entirely new source :)

I'm just pulling from my fork ATM, so no stress.

PR #205, opened 2020-07-20
Add mime-support as upstream source for MIME types.

sgpinkus:
File structure is identical to Apache as the upstream file is a drop in replacement for the Apache mime.types files (actually it's the other way around ..).

In this PR I've added a basic script to print out how many MIME from each source. I've also labelled custom types explicitly to help with tracing source in `db.json`:

**Before:**

    total 2193
    iana 1861
    apache 250
    nginx 13
    mime-support 0
    custom 0
    undefined 0
    other 69

**After:**

    total 2381
    iana 1861
    apache 250
    nginx 13
    mime-support 192
    custom 65
    undefined 0
    other 0

---

 - *Copy right info:* https://salsa.debian.org/debian/mime-support/-/blob/master/debian/copyright
 - *DB Lineage:* A major upstream source of these MIME types is Freedesktop.org [shared-mime-info](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in). Beyond that it's unclear exactly but you can see the commit history [here](https://salsa.debian.org/debian/mime-support/-/commits/master/mime.types). I feel it is much the same as Apache and Nginx: The maintainers just add entries as they see fit when it's reasonable. Also note, it is this file that is the default mime.types file, and the file that will be used by Apache server (or probably any other kind of HTTP server) running on a Debian based distro, given Apache is installed from pkg manager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,10 @@ The JSON file is a map lookup for lowercased mime types.
 Each mime type has the following properties:
 
 - `.source` - where the mime type is defined.
-    If not set, it's probably a custom media type.
     - `apache` - [Apache common media types](http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types)
     - `iana` - [IANA-defined media types](http://www.iana.org/assignments/media-types/media-types.xhtml)
     - `nginx` - [nginx media types](http://hg.nginx.org/nginx/raw-file/default/conf/mime.types)
+    - `custom` - Loaded from custom set of types maintained in this repository.
 - `.extensions[]` - known extensions associated with this mime type.
 - `.compressible` - whether a file of this type can be gzipped.
 - `.charset` - the default charset associated with this type, if any.
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Each mime type has the following properties:
     - `apache` - [Apache common media types](http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types)
     - `iana` - [IANA-defined media types](http://www.iana.org/assignments/media-types/media-types.xhtml)
     - `nginx` - [nginx media types](http://hg.nginx.org/nginx/raw-file/default/conf/mime.types)
+    - `mime-support` [Debian Linux mime.types database (upstream source)](https://salsa.debian.org/debian/mime-support/-/raw/master/mime.types)
     - `custom` - Loaded from custom set of types maintained in this repository.
 - `.extensions[]` - known extensions associated with this mime type.
 - `.compressible` - whether a file of this type can be gzipped.
--- a/db.json
+++ b/db.json
@@ -76,6 +76,10 @@
     "source": "iana",
     "extensions": ["ez"]
   },
+  "application/annodex": {
+    "source": "mime-support",
+    "extensions": ["anx"]
+  },
   "application/applefile": {
     "source": "iana"
   },
@@ -106,6 +110,11 @@
   },
   "application/atomicmail": {
     "source": "iana"
+  },
+  "application/atomserv+xml": {
+    "source": "mime-support",
+    "compressible": true,
+    "extensions": ["atomsrv"]
   },
   "application/atomsvc+xml": {
     "source": "iana",
@@ -148,7 +157,12 @@
   "application/batch-smtp": {
     "source": "iana"
   },
+  "application/bbolin": {
+    "source": "mime-support",
+    "extensions": ["lin"]
+  },
   "application/bdoc": {
+    "source": "custom",
     "compressible": false,
     "extensions": ["bdoc"]
   },
@@ -309,6 +323,7 @@
     "source": "iana"
   },
   "application/dart": {
+    "source": "custom",
     "compressible": true
   },
   "application/dash+xml": {
@@ -338,7 +353,8 @@
     "compressible": true
   },
   "application/dicom": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["dcm"]
   },
   "application/dicom+json": {
     "source": "iana",
@@ -375,6 +391,10 @@
   "application/dskpp+xml": {
     "source": "iana",
     "compressible": true
+  },
+  "application/dsptype": {
+    "source": "mime-support",
+    "extensions": ["tsp"]
   },
   "application/dssc+der": {
     "source": "iana",
@@ -493,6 +513,7 @@
     "compressible": true
   },
   "application/fido.trusted-apps+json": {
+    "source": "custom",
     "compressible": true
   },
   "application/fits": {
@@ -502,7 +523,8 @@
     "source": "iana"
   },
   "application/font-sfnt": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["otf","ttf"]
   },
   "application/font-tdpfr": {
     "source": "iana",
@@ -510,11 +532,16 @@
   },
   "application/font-woff": {
     "source": "iana",
-    "compressible": false
+    "compressible": false,
+    "extensions": ["woff"]
   },
   "application/framework-attributes+xml": {
     "source": "iana",
     "compressible": true
+  },
+  "application/futuresplash": {
+    "source": "mime-support",
+    "extensions": ["spl"]
   },
   "application/geo+json": {
     "source": "iana",
@@ -530,6 +557,9 @@
   "application/geoxacml+xml": {
     "source": "iana",
     "compressible": true
+  },
+  "application/ghostview": {
+    "source": "mime-support"
   },
   "application/gltf-buffer": {
     "source": "iana"
@@ -561,7 +591,12 @@
     "compressible": true
   },
   "application/hjson": {
+    "source": "custom",
     "extensions": ["hjson"]
+  },
+  "application/hta": {
+    "source": "mime-support",
+    "extensions": ["hta"]
   },
   "application/http": {
     "source": "iana"
@@ -677,6 +712,7 @@
     "source": "iana"
   },
   "application/json5": {
+    "source": "custom",
     "extensions": ["json5"]
   },
   "application/jsonml+json": {
@@ -736,6 +772,10 @@
   "application/lxf": {
     "source": "iana"
   },
+  "application/m3g": {
+    "source": "mime-support",
+    "extensions": ["m3g"]
+  },
   "application/mac-binhex40": {
     "source": "iana",
     "extensions": ["hqx"]
@@ -753,6 +793,7 @@
     "extensions": ["mads"]
   },
   "application/manifest+json": {
+    "source": "custom",
     "charset": "UTF-8",
     "compressible": true,
     "extensions": ["webmanifest"]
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -49,10 +49,11 @@ The JSON file is a map lookup for lowercased mime types.
 Each mime type has the following properties:
 
 - `.source` - where the mime type is defined.
-    If not set, it's probably a custom media type.
     - `apache` - [Apache common media types](http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types)
     - `iana` - [IANA-defined media types](http://www.iana.org/assignments/media-types/media-types.xhtml)
     - `nginx` - [nginx media types](http://hg.nginx.org/nginx/raw-file/default/conf/mime.types)
+    - `mime-support` [Debian Linux mime.types database (upstream source)](https://salsa.debian.org/debian/mime-support/-/raw/master/mime.types)
+    - `custom` - Loaded from custom set of types maintained in this repository.
```
dougwilson, line 56: Please revert this inside this PR and if you think calling this out is worthwhile, please open a PR or issue specifically about this. We don't currently because the JSON file is already quite large, so keeping one of the source to blank really helps folks who are sending this JSON to web clients.

scripts/fetch-mime-support.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+var getMimeTypes = require('./fetch-mime-types-file');
+
+/**
+ * URL for the mime.types file in the Apache HTTPD project source.
+ */
+var URL = 'https://salsa.debian.org/debian/mime-support/-/raw/master/mime.types?inline=false'
```
dougwilson, line 7: Please make sure to follow up here with the permission of the site operator that we can begin to start polling / scraping their endpoint with an automated process. I couldn't find any public TOS on a quick look, so if there is one that says it's OK, then that's fine and we don't need their explicit permission.

scripts/stats.js
```diff
@@ -0,0 +1,8 @@
+const db = require('../db.json');
```
dougwilson, line 1: I would suggest moving this into a separate PR, as it's a separate change from adding the mime-support upstream.

db.json
```diff
@@ -1066,7 +1117,8 @@
     "extensions": ["pgp"]
   },
   "application/pgp-keys": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["key"]
```
dougwilson, line 1121: Where does this come from? The official record for this type (https://tools.ietf.org/html/rfc3156#section-9.3) states the file extension is ".asc". It jumped out to me since ".key" is the Apple Keynote files.

db.json
```diff
@@ -768,7 +809,7 @@
   },
   "application/mathematica": {
     "source": "iana",
-    "extensions": ["ma","nb","mb"]
+    "extensions": ["ma","nb","mb","nbp"]
```
dougwilson, line 812: Where does this new extension come from? I see there is a registration from Wolfram Research (the owners of the spec) at https://www.iana.org/assignments/media-types/application/mathematica but this extension is not listed there.

db.json
```diff
@@ -502,20 +523,26 @@
...
   },
   "application/font-tdpfr": {
     "source": "iana",
     "extensions": ["pfr"]
   },
   "application/font-woff": {
     "source": "iana",
-    "compressible": false
+    "compressible": false,
+    "extensions": ["woff"]
```
dougwilson, line 536: This mapping is obsolete; many of these font types (like .woff) moved under font/ tree (as font/woff in this case). More information can be found here: http://tools.ietf.org/rfc/rfc8081.txt

db.json
```diff
@@ -502,20 +523,26 @@
     "source": "iana"
   },
   "application/font-sfnt": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["otf","ttf"]
```
dougwilson, line 527: This mapping is obsolete; many of these font types (moved under font/ tree. More information can be found here: http://tools.ietf.org/rfc/rfc8081.txt

db.json
```diff
@@ -2351,7 +2419,8 @@
     "source": "iana"
   },
   "application/vnd.debian.binary-package": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["deb","ddeb","udeb"]
```
dougwilson, line 2423: I see deb and udeb listed in the type registration https://www.iana.org/assignments/media-types/application/vnd.debian.binary-package , but where does ddeb come from?

db.json
```diff
@@ -5680,17 +5909,27 @@
...
+  },
   "application/x-mobipocket-ebook": {
     "source": "apache",
     "extensions": ["prc","mobi"]
   },
   "application/x-mpegurl": {
-    "compressible": false
+    "source": "mime-support",
+    "compressible": false,
+    "extensions": ["m3u8"]
```
dougwilson, line 5923: It was my understanding this extension is for the type "application/vnd.apple.mpegurl", right? Looking at the spec https://tools.ietf.org/html/rfc8216#section-4 the section says:

>    Each Playlist file MUST be identifiable either by the path component
   of its URI or by HTTP Content-Type.  In the first case, the path MUST
   end with either .m3u8 or .m3u.  In the second, the HTTP Content-Type
   MUST be "application/vnd.apple.mpegurl" or "audio/mpegurl".  Clients
   SHOULD refuse to parse Playlists that are not so identified.

README.md
sgpinkus: OK. Done.

scripts/stats.js
sgpinkus: Done.

db.json
```diff
@@ -7418,7 +8024,7 @@
   },
   "text/calendar": {
     "source": "iana",
-    "extensions": ["ics","ifb"]
+    "extensions": ["ics","ifb","icz"]
```
dougwilson, line 8027: I see in the text/calendar spec https://tools.ietf.org/html/rfc5545#section-8.1 section about the type, it lists the ics and ifb file extensions, but no mention of the icz extension.

db.json
```diff
@@ -6750,15 +7289,16 @@
...
   "font/sfnt": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["ttf","otf"]
   },
   "font/ttf": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["ttf"]
+    "extensions": ["ttf","otf"]
```
dougwilson, line 7301: Why add .otf to this? My understanding is that .otf = font/otf and .ttf = font/ttf (which is what the database was prior to adding this source).

scripts/fetch-mime-support.js
sgpinkus: Yes agree required for merge. Will do.

db.json
sgpinkus: gnupg uses this for keys.

db.json
sgpinkus: Obsolete yes. Is not being obsolete prerequisite for the extension inclusion? Many files are going to have that extension regardless of whether someone declares is obsolete.

db.json
sgpinkus: Obsolete yes. Is not being obsolete prerequisite for the extension inclusion? Many files are going to have that extension regardless of whether someone declares is obsolete.

db.json
dougwilson: Gotcha. Can you link to the source of that information?

db.json
sgpinkus: The mapping is not reversible. They both have the same extension?

db.json
dougwilson: The mime is what is obsolete, not the extension. The files would simply use the non-obsolete file extension. My comment is about mapping the extension here to the obsolete type instead of the current type.

db.json
dougwilson: The mime is what is obsolete, not the extension. The files would simply use the non-obsolete file extension. My comment is about mapping the extension here to the obsolete type instead of the current type.

db.json
dougwilson: I'm not sure what your comment means, I'm sorry. Can you state it a different way?

sgpinkus: File with MIME type `application/x-mpegurl` and `application/vnd.apple.mpegurl` both can have ext m3u8.

dougwilson: Yep, I get it. My question is can you site where the MIME type "application/x-mpegurl" comes from? I don't see it in the specification anywhere. Without a source, one could always argue that "foo/bar" MIME type is also m3u8 :)

sgpinkus: Not sure. To answer all of your questions, which are legit, will have to get more info on how this DB was pulled together, from the maintained. Don't expect you to merge with out that. Fact remains though this the Debian Linux `/etc/mime.types` file source, so in my mind that gives it as much legitimacy as the Apache or Nginx versions.

dougwilson:
@sgpinkus I'm looking into what is coming in because there are several known places that use this module to provide file extension -> mime type mappings: The Node.js package https://www.npmjs.com/package/mime and GitHub Pages https://docs.github.com/en/enterprise/2.13/user/articles/mime-types-on-github-pages are two very notable ones. So we have to be very careful pulling in conflicting entries and understand how they are going to affect the behavior of these projects if there are conflicting mappings pulled in.

db.json
dougwilson: Sure, but that doesn't mean they do not have outdated or invalid entries. When we pulled in NGINX and Apache back in the day, we did this same process and fixed a lot of bad data. That is what we'd need to do in this same case.

sgpinkus:
> @sgpinkus I'm looking into what is coming in because there are several known places that use this module to provide file extension -> mime type mappings ...

OK. Understood. Still waiting on more info from maintainer. Will get back to you with any new info. If you can't merge that is totally fine.

db.json
sgpinkus: Ah OK . That clears it up. Not sure if I have the time right now for all that!

sgpinkus:
@dougwilson I got a reply from the maintainer of mime-support, Charles Plessy. I seems many of the MIME types in mime-support DB were added over the years by hand. He indicated that pulling directly from IANA would have been preferable now that IANA has become more recieptive to adding new MIME types.

I pointed out that you have a script that pulls from IANA already. And directed him to this repository, and also this PR.

Still it would be useful to get some of the types in mime-support added here. I guess, in retrospect, the proper process for doing this is actually, 1. IANA, 2. pre IANA the adhoc "custom" type registration process you have set up here.

---

For reference, [here](https://gist.github.com/sgpinkus/6c4084584e1542e35c80fe61195814e4) are the 192 MIME types in mime-support not in mime-db, in CSV, and JSON. Script to used generate is included.

younggun23:
https://github.com/younggun23/mime-db/Add mime-support as upstream source for MIME types

README.md
```diff
@@ -49,10 +49,11 @@ The JSON file is a map lookup for lowercased mime types.
...
 - `.extensions[]` - known extensions associated with this mime type.
```
NaingLinnKyaw, line 57: ``

jayvdb:
Is this still being worked on?

PR #217, opened 2020-10-11
Create npm-publish.yml

AmEyOs071:


dougwilson:
Thank you, unfortunately automating publishing does not work with 2FA-protected packages like this one.

PR #218, opened 2020-10-20
fix: add .opus as valid extension for audio/ogg

### LINKED ISSUE #193: ".opus" extension for audio/opus
`audio/opus` type should map to `.opus` extension.

https://en.wikipedia.org/wiki/Opus_(audio_format)

marc0s:
As per https://www.iana.org/assignments/media-types/audio/ogg .opus
should be considered a valid extension for `audio/ogg` mediatype.

Fixes #193

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Add extension `.opus` to `audio/ogg`
   * Add new upstream MIME types
 
 1.45.0 / 2020-09-22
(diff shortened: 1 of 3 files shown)


dougwilson (changes requested):
The db.json file cannot be edited directly, as it is a generated file. Please follow contribution instructions https://github.com/jshttp/mime-db#contributing

marc0s:
Thanks for the pointers, missed that.

I did the changes and added the `opus` extension to the `src/custom-types.json` file. Don't know why the other extensions are on the apache file, though.

dougwilson (commented):
thanks!

PR #221, opened 2020-11-06
feat: add `text/x-java-properties` mime type

### LINKED ISSUE #220: Missing mime type `text/x-java-properties`
There is missing `text/x-java-properties` for `*.properties` files.

k3nsei:
This PR closes #220

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -805,7 +805,11 @@
   },
   "text/x-java-properties": {
     "compressible": true,
-    "extensions": ["properties"]
+    "extensions": ["properties"],
+    "sources": [
+      "https://stackoverflow.com/a/28813973",
+      "https://www.digipres.org/formats/mime-types/"
+    ]
   },
   "text/x-jquery-tmpl": {
     "compressible": true
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -803,6 +803,9 @@
   "text/x-handlebars-template": {
     "extensions": ["hbs"]
   },
+  "text/x-java-properties": {
+    "extensions": ["properties"]
+  },
   "text/x-jquery-tmpl": {
     "compressible": true
   },
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -804,6 +804,7 @@
     "extensions": ["hbs"]
   },
   "text/x-java-properties": {
+    "compressible": true,
     "extensions": ["properties"]
   },
   "text/x-jquery-tmpl": {


src/custom-types.json
```diff
@@ -803,6 +803,10 @@
   "text/x-handlebars-template": {
     "extensions": ["hbs"]
   },
+  "text/x-java-properties": {
```
dougwilson, line 806: Please add a sources key and include a primary source for the type data (https://github.com/jshttp/mime-db#contributing).

dougwilson (changes requested):

src/custom-types.json
k3nsei: Done

src/custom-types.json
```diff
@@ -805,7 +805,11 @@
   },
   "text/x-java-properties": {
     "compressible": true,
-    "extensions": ["properties"]
+    "extensions": ["properties"],
+    "sources": [
+      "https://stackoverflow.com/a/28813973",
```
dougwilson, line 810: Thank you. These links need to be a [primary source](https://en.wikipedia.org/wiki/Primary_source) (https://github.com/jshttp/mime-db#contributing).

dougwilson (changes requested):

src/custom-types.json
```diff
@@ -803,6 +803,14 @@
   "text/x-handlebars-template": {
     "extensions": ["hbs"]
   },
+  "text/x-java-properties": {
+    "compressible": true,
+    "extensions": ["properties"],
+    "sources": [
+      "https://stackoverflow.com/a/28813973",
+      "https://www.digipres.org/formats/mime-types/"
```
dougwilson, line 811: https://github.com/jshttp/mime-db#contributing

> links to type aggregating sites [...] are not acceptable

k3nsei: I think this will be https://trac.edgewall.org/wiki/TracSyntaxColoring and http://wiki.cacert.org/HelpOnParsers

k3nsei: Are those valid primary sources?

dougwilson: https://trac.edgewall.org/wiki/TracSyntaxColoring does not look like a primary source, as they did not create this type/extension and whatever the spec for it's contents; it is just listing information for how their editor works.

http://wiki.cacert.org/HelpOnParsers does not look like a primary source, as they did not create this type/extension and whatever the spec for it's contents; it is just listing information for how their parser works.

These would be considered secondary sources; they would have gotten that information from some primary location.

k3nsei: I can't find any source that can be considered primary. But this is commonly used mime type for `.properties` files

dougwilson: Thank you. Without a primary source, we cannot just add it here directly. You should follow the process at https://github.com/jshttp/mime-db#adding-custom-media-types

Issue #86, opened 2017-07-14
video/quicktime should map to .mov, not .qt


timcosta:
Running into an issue where files with the extension `.mov` are getting the mimetype `video/qucktime`, and then if we decode the mimetype it comes back as `.qt`.

You can see that apache has updated their spec here: https://issues.apache.org/jira/browse/TIKA-1882

dougwilson:
It is mapped to both in our database, as the type could be either or: https://github.com/jshttp/mime-db/blob/be37b05719ad488f4dcd5fd2c901f59608aeff1e/db.json#L6661-L6665

Apache also maps the type to both extensions in their mapping (http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types):

> video/quicktime					qt mov

So what is the issue?

timcosta:
@dougwilson I suppose my issue is likely in `node-mime` then, as it is returning `.qt` and not `.mov` for `video/quicktime`. Sorry about that, and thanks for the fast response!

dougwilson:
Gotcha. I looked though out sources (we really just aggregate Apache + nginx + IANA) and it's just the Apache mappings that are providing the `.qt` extension mapping. If the Apache projects removes it, then it'll automatically get removed from this module, as the nature that we are just an aggregation. Our format is just a list of whatever extensions our sources have on file for a given type in an array.

I hope that helps / makes sense :) !

aitchkhan:
Can we change the order and keep it like the following? Because the default for `video/quicktime` is `mov`. 
Even when I try to rename a file with `mov` to `qt` for saving a file from the internet. Mozilla Firefox asks me, do you really want to use `qt`? The standard is `mov`.

PS: I am not sure about the impact of this.

https://github.com/jshttp/mime-db/blob/be37b05719ad488f4dcd5fd2c901f59608aeff1e/db.json#L6661-L6665

```
  "video/quicktime": {
    "source": "apache",
    "compressible": false,
    "extensions": ["mov", "qt"]
  }
```

The packages dependent on mime-db returns the first element in the list, which in our case will return `qt`.

Screenshot of Mozilla Firefox:
<img width="858" alt="Screen Shot 2020-12-02 at 10 09 14 AM" src="https://user-images.githubusercontent.com/8606435/100830936-91b8c700-3486-11eb-9a52-f0931909cd1b.png">

dougwilson:
> Can we change the order and keep it like the following?

[...]

PR #223, opened 2020-12-02
update db.json and change order of video/quicktime

aitchkhan:
Trying to fix https://github.com/jshttp/mime-db/issues/86

### DIFF
--- a/db.json
+++ b/db.json
@@ -8033,7 +8033,7 @@
   "video/quicktime": {
     "source": "iana",
     "compressible": false,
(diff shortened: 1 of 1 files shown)


dougwilson (changes requested):
The db.json file is a generated file and cannot be changed. Please see readme for contribution instructions.

dougwilson:
Based on #86 that ordering is coming from the Apache project, which is where you would need to make you fix to and then we can pull in the changes from them.

PR #225, opened 2020-12-08
Remove charset from application/json

rtm516:
This was done to fix Azure domain registration with GitHub pages.
It has had multiple issues in multiple repositories and this seems like the only way to fix it.

https://docs.microsoft.com/en-us/answers/questions/13291/publisher-domain-verification-fails-because-verifi.html
https://github.com/MicrosoftDocs/azure-docs/issues/46080
https://github.com/jekyll/jekyll/issues/8006

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -71,7 +71,6 @@
     ]
   },
   "application/json": {
-    "charset": "UTF-8",
     "compressible": true,
     "extensions": ["map"]
   },


dougwilson (commented):
Hi, I'm not sure what you are trying to achieve,  but this is simply a database of information. And this piece of information seems correct to me:

> .charset - the default charset associated with this type, if any.

The spec at https://tools.ietf.org/html/rfc8259#section-8.1 seems to indicate that utf-8 is the default. Especially when this mime type in the wild often does not include a charset parameter, it is important that this database store what the default charset is so folks can do a lookup.

dougwilson:
After reading the links you provided, I don't see how anything there relates to this module. Due to that, and from what I mentioned in the previous comment, I'm going to close this for now unless there is additional information that comes to light.

PR #226, opened 2020-12-14
Add SHTM support

boemekeld:


dougwilson (changes requested):
Hi, and thank you for the pull request. The `db.json` file is actually auto generated and cannot be changed by itself through a pull request. You can find the instructions for contributing in the readme here: https://github.com/jshttp/mime-db#contributing

dougwilson (changes requested):
Please make sure to include the source for where the MIME type and the associated extension is sourced from. This needs to be a primary source; links to type aggregating sites and Wikipedia are not acceptable. Once we validate the source of the information, your PR will be merged. Thank you so much for contributing directly to our database!

boemekeld:
Hi @dougwilson, thanks for answering.
Is it correct now?

dougwilson:
Hi @boemekeld please make sure to include the source for where the MIME type and the associated extension is sourced from. This needs to be a primary source; links to type aggregating sites and Wikipedia are not acceptable. Once we validate the source of the information, your PR will be merged. Thank you so much for contributing directly to our database!

boemekeld:
@dougwilson SHTM is a common extension for Server-Side Includes, it is less used than SHTML, but it is present in thousands of pages.

It is possible to find references to this extension in some documentation. I believe that its origin comes from Microsoft, but this is not documented.

Here are some links, I hope it will be enough.
https://en.wikipedia.org/wiki/Server_Side_Includes
http://erlang.org/documentation/doc-4.7.3/lib/inets-2.2.5/doc/html/mod_include.html

Thank you.

dougwilson:
Thank you for that, unfortunately neither of those links are sufficient. This module provides a json database of type information sourced from the three places listed at the top of the readme.

Ideally that is where the addition should be made.

But we do allow folks to contribute directly here, bypassing the aggregation. To do so, though, we need sufficient evidence that the given mime type is associated with a given file extension. Typically the folks who create a given mime type will list out what file extensions are reserved to hold that data.

Wikipedia is not a primary source of information, so we don't accept links to Wikipedia; please link directly to where wikipedia sourced that piece of information that is relevant.

The second link you included does not seem to say anything about the mime type "text/html", so I am not sure how that page provides a primary source of information that says the .shtm file extension is associated to the mime type text/html.

boemekeld:
I found this reference in the Microsoft IIS 6 documentation.
https://docs.microsoft.com/en-us/previous-versions/iis/6.0-sdk/ms525105(v=vs.90)

dougwilson:
I'm not sure if you copied the correct link or not, but that page does not seem to say anything about the mime type "text/html", so I am not sure how that page provides a primary source of information that says the .shtm file extension is associated to the mime type text/html.

boemekeld:
We are talking about a very old file extension. Probably released by IIS 4 or 5. So it is difficult to find documentation.

You are right, this link does not mention the mime type "text/html", however, it relates to SHTML, which is also "text/html".

"By default, the file name extensions .stm, .shtm, and .shtml are mapped to the SSI interpreter"

dougwilson:
I have seen many before say "trust me" and then when merged, more come and say it is wrong and say "trust me". Our policy is that we need concrete evidence of this association to include it here. The three places we source from may have different standards for inclusion if you would rather pursue their policies for inclusion, which we would pull in automatically any updates from those sources.

boemekeld:
@dougwilson I understand your decision and I don't want to take your time anymore.

I regret that this cannot be understood as a primary source. It would help my project and certainly others.

I will make one last attempt, with 3 links (including an Apache link!), In all of them, SHTM and SHTML are treated the same way.

If you don't understand it as a primary source, feel free to close this pull request.

Thank's for your time.

https://cwiki.apache.org/confluence/display/HTTPD/ApacheVirtualHostMysql+hosts.conf
https://docs.microsoft.com/en-us/previous-versions/office/sharepoint-foundation-2010/cc287701(v=office.14)
https://docs.microsoft.com/en-us/previous-versions/aa711476(v=vs.71)

dougwilson:
Thank you for those. I would suggest that you go through the IANA process linked on the readme, where they can help you better. We already source from Apache's file extension to mime mapping.

PR #227, opened 2020-12-15
.toml as text/plain

Heteroculturalism:
Stellar exchanges need to read stellar.toml files as text/plain (see https://developers.stellar.org/docs/issuing-assets/publishing-asset-info/).

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -761,7 +761,7 @@
   },
   "text/plain": {
     "compressible": true,
-    "extensions": ["ini"]
+    "extensions": ["ini", "toml"]
   },
   "text/richtext": {
     "compressible": true


dougwilson (changes requested):
Hello, and thank you for contributing directly to our database. Please be sure to add a primary source for your mapping according to our contributing instructions (https://github.com/jshttp/mime-db#contributing).

I see you provided a link in your PR body, however, this is not a primary source for either .toml nor for text/plain. It seems to be some instructions for how to get a web browser to display data inline in a web browser, which the same instructions of "send `content-type: text/plain` would apply for any file extension. This database is regarding the true, actual mime type associations of the content.

I found the official specification / site for TOML and it appears that the correct type of `application/toml`, not `text/plain`. https://toml.io/en/v1.0.0-rc.3#mime-type

dougwilson:
It looks like the .toml association was added in version 1.41.0 of the database (https://github.com/jshttp/mime-db/blob/master/HISTORY.md#1410--2019-08-30)

PR #231, opened 2021-03-11
Add extensions for the "application/tar" type

zhanglin-wu:
Extensions like "tgz" is commonly used, for example, the "[npm pack](https://docs.npmjs.com/cli/v7/commands/npm-pack)" command generates a .tgz archive.

But these extensions related to the "application/tar" type is not included by mime-db yet, so let's add them here.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -142,7 +142,12 @@
     ]
   },
   "application/tar": {
-    "compressible": true
+    "compressible": true,
+    "extensions": ["tar","tar.gz","taz","tgz"],
+    "sources": [
+      "https://www.gnu.org/software/tar/",
+      "https://en.wikipedia.org/wiki/Tar_(computing)"
+    ]
   },
   "application/toml": {
     "extensions": ["toml"],
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -145,8 +145,7 @@
     "compressible": true,
     "extensions": ["tar","tar.gz","taz","tgz"],
     "sources": [
-      "https://www.gnu.org/software/tar/",
-      "https://en.wikipedia.org/wiki/Tar_(computing)"
+      "https://www.gnu.org/software/tar/"
     ]
   },
   "application/toml": {


zhanglin-wu:
We open the source links to see more details about the extensions related to the "application/tar" type:

- https://www.gnu.org/software/tar/
- https://en.wikipedia.org/wiki/Tar_(computing)

src/custom-types.json
```diff
@@ -142,7 +142,12 @@
     ]
   },
   "application/tar": {
-    "compressible": true
+    "compressible": true,
+    "extensions": ["tar","tar.gz","taz","tgz"],
+    "sources": [
+      "https://www.gnu.org/software/tar/",
```
dougwilson, line 148: I cannot find where on this page any of these extensions are listed as being for this mime type. If you can point it out, I maye have overlooked.

src/custom-types.json
```diff
@@ -142,7 +142,12 @@
...
+      "https://en.wikipedia.org/wiki/Tar_(computing)"
```
dougwilson, line 149: Links to wikipedia are not accepted as per the readme, you can remove this.

dougwilson (changes requested):

src/custom-types.json
zhanglin-wu: Is this page link sufficient enough? I can change to this if it is.
https://www.gnu.org/software/tar/manual/html_node/gzip.html

The page contains a list of extensions like this:
```
Suffix	Compression program
‘.gz’	gzip
‘.tgz’	gzip
‘.taz’	gzip
‘.Z’	compress
‘.taZ’	compress
‘.bz2’	bzip2
‘.tz2’	bzip2
‘.tbz2’	bzip2
‘.tbz’	bzip2
‘.lz’	lzip
‘.lzma’	lzma
‘.tlz’	lzma
‘.lzo’	lzop
‘.xz’	xz
‘.zst’	zstd
‘.tzst’	zstd
```

dougwilson: The page needs to say that a given extension is identified by a given mime type. Is there a page that procides that or a page with a table of those archive formats to their mime type?

src/custom-types.json
zhanglin-wu: Done. Thank you!

src/custom-types.json
zhanglin-wu: Can't find such a page unfortunately. There is no official definition on the ".tgz" extension. This is a discussion that might be helpful on coming up with a mime type for the extension:
https://superuser.com/questions/901962/what-is-the-correct-mime-type-for-a-tar-gz-file

Can we define a type for it which we think appropriate for now? And then if there is an official definition, we can switch to it? I think either of these two types is acceptable for me:
- application/tar
- application/tar+gzip

I'm open to other ideas as well.

dougwilson: Thia module is just an aggregatation of the three sources listed at the top of the readme. As a curiosity if none of those three sources have the mapping but there is a clear, official source, we can add it here. Absent that, we wouldn't add it.

zhanglin-wu: Just to double check that I understand your comment. Are you saying that if we don't provide a clear and official web page, you will not approve this Pull Request?

dougwilson: The requirements are listed at https://github.com/jshttp/mime-db#contributing

zhanglin-wu: Thank you! I will try to follow the requirements and see if we get this done.
In the meantime, closing this PR. Will re-open it if needed.

PR #233, opened 2021-03-19
Update db.json

clearwinggames:
Adding ghtml for (cached) generated games embedded in an html page using HTML5 Canvas


dougwilson:
Hi @clearwinggames thanks for the PR! You cannot make PRs against the `db.json` file as it is autogenerated. You can read more about making PRs in our README: https://github.com/jshttp/mime-db#contributing

PR #236, opened 2021-05-12
More MapBox Vector Tile info

tannewt:
From: https://www.iana.org/assignments/media-types/application/vnd.mapbox-vector-tile

This will allow compression of the files when statically hosted.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Add extension `.mvt` to `application/vnd.mapbox-vector-tile`
   * Add new upstream MIME types
 
 1.47.0 / 2021-04-01
--- a/db.json
+++ b/db.json
@@ -3476,7 +3476,8 @@
     "extensions": ["portpkg"]
   },
   "application/vnd.mapbox-vector-tile": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["mvt"]
   },
   "application/vnd.marlin.drm.actiontoken+xml": {
     "source": "iana",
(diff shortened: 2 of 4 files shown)


dougwilson (changes requested):
Hello, and thank you for your contribution! Please make sure to follow our contribution guidelines at https://github.com/jshttp/mime-db#contributing

Notably, you have made modifications to the `db.json` file, which is a generated file and the changes will be lost. If you have any questions that the contributing guide does not answer for how to make a pull request, please let me know so I can help.

1. Please move your modifications to the appropriate src/ file.
2. Please add the link you provided in the pull request into the entry you create in the file.
3. Please additionally add a link for how the compressibility was determined (but please be aware that in the next major version of this database we will no longer carry compressibility data).

tannewt (commented):
Thanks for the fast reply! Sorry I missed the source files. I know it's compressible because the link states it's protocol buffers which are. (It is commonly used compressed too but I don't know of an "official" source for that.)

dougwilson:
Hi @tannewt thanks for updating the PR. Regarding the compressible flag, from what I understand of protobuf is that it is simply a serialization format, that itself is not compressible (binary). But the contents that are being serialized inside that container may make it compressible within for example strings or byte arrays. Can you provide some specific examples of what this file looks like that demonstrates that it is compressible? Unfortunately the answer of "I know it's compressible" is no longer acceptable as folks come here to then argue about what we have in the database. We need some kind of source to link to that speaks on this data point.

tannewt:
Thanks @dougwilson. Could you give me an example of a url that satisfies this? I just did compression on some local files I have and they went from 717M to 607M.

dougwilson:
I can't speak for the type you are proposing, but as an example here is what was provided recently for some font types: http://www.phpied.com/gzip-your-font-face-files/

tannewt:
Thanks! That's what I was hoping for. Unfortunately, I found https://wiki.openstreetmap.org/wiki/PBF_Format which states individual blocks may be compressed. So, I've removed it from here. (I'm not sure why the extension changed from mvt to pbf but it's the same format.)

dougwilson (approved):

PR #239, opened 2021-06-15
Add file extensions for video/mp2t type

pniewiejski:
While using https://github.com/jshttp/mime-types I noticed that it fails to perform lookup for `*.m2ts` files. Once this PR gets merged, published, and `mime-types` is updated to use the updated version of mime-db the problem should be resolved.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -858,6 +858,9 @@
   "video/mpeg": {
     "compressible": false
   },
+  "video/mp2t": {
+    "extensions": ["tsv", "tsa", "m2t", "m2ts"]
+  },
   "video/ogg": {
     "compressible": false
   },


dougwilson (changes requested):
https://github.com/jshttp/mime-db#contributing

Please include link(s) to the primary sources where the file extensions are associated to the mime type.

pniewiejski:
Thanks for quick response. AFAIK This mime type was originally introduced in [RFC 3555](https://tools.ietf.org/rfc/rfc3555.txt) and it’s also listed by iana (https://www.iana.org/assignments/media-types/video/MP2T) but they **do not** explicitly name `“tsv”, “tsa”, “m2t”, “m2ts”` as extensions, even though they are commonly used extensions.

I'll try to look up they are not mentioned in some other documentation.

pniewiejski:
Hi,
Sorry for taking such a long time. I did look for a meaningful documentation that links these formats with `video/mp2t` but I only found a reference in Blu-ray. I thought that it would be easier to find some official documentation given that this mime type is commonly used with these extensions. I'm going to close this pull request.

* `*.m2ts`
    * Reference: Blue-Ray Disc [Audio Visual Application Format Specifications for BD-ROM](http://www.blu-raydisc.com/Assets/Downloadablefile/BD-ROM_Part3_V3.2_WhitePaper_180122(clean).pdf)
    * This file extension is used for BDAV MPEG-2 Transport Stream which is an extension of MPEG-2 Transport Stream
      used for Blue-Ray discs.
      > The “zzzzz.m2ts” files contains a BDAV MPEG-2 transport stream.

PR #240, opened 2021-06-28
Add PY ext

syscaday:
Would like to add for PY ext

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -832,10 +832,16 @@
     ]
   },
   "text/x-python-script": {
-    "extensions": ["py"]
+    "extensions": ["py"],
+    "sources": [
+      "http://www.htmlcodes.ws/reference/mime-types-complete-list.cfm"
+    ]
   },
   "text/x-python": {
-    "extensions": ["py"]
+    "extensions": ["py"],
+    "sources": [
+      "https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=235663#5"
+    ]
   },
   "text/x-sass": {
     "extensions": ["sass"]
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -831,6 +831,12 @@
       "https://en.wikipedia.org/wiki/Processing_(programming_language)"
     ]
   },
+  "text/x-python-script": {
+    "extensions": ["py"]
+  },
+  "text/x-python": {
+    "extensions": ["py"]
+  },
   "text/x-sass": {
     "extensions": ["sass"]
   },


dougwilson (changes requested):
Thank you! Please include an array of URLs of where the MIME type and the associated extensions are sourced from. This needs to be a primary source; links to type aggregating sites and Wikipedia are not acceptable.

syscaday:
text/x-python
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=235663#5

syscaday:
text/x-python-script == text/x-python.script
https://www.sitepoint.com/mime-types-complete-list/

in ubuntu server capture as - not .

dougwilson (changes requested):
Please include an array of URLs of where the MIME type and the associated extensions are sourced from. This needs to be a primary source; links to type aggregating sites and Wikipedia are not acceptable.

syscaday:
Added source URL

src/custom-types.json
```diff
@@ -831,6 +831,18 @@
       "https://en.wikipedia.org/wiki/Processing_(programming_language)"
     ]
   },
+  "text/x-python-script": {
+    "extensions": ["py"],
+    "sources": [
+      "http://www.htmlcodes.ws/reference/mime-types-complete-list.cfm"
```
dougwilson, line 837: This URL does not contain a reference for the MIME type entered `text/x-python-script`.

src/custom-types.json
```diff
@@ -831,6 +831,18 @@
...
+    ]
+  },
+  "text/x-python": {
+    "extensions": ["py"],
+    "sources": [
+      "https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=235663#5"
```
dougwilson, line 843: > The mime types for python files, according to the python cvs, 
are:


Please link to the primary source, not a secondary source.

dougwilson (changes requested):

src/custom-types.json
Martii: @dougwilson 

Not sure if you will take debian source still as a primary source but several are at:
* https://sources.debian.org/src/media-types/4.0.0/mime.types/#L1513
* https://sources.debian.org/src/media-types/4.0.0/mime.types/#L2073  *(more probable what this PR is for)*

... since they are `x-` prefixed there's a strong possibility they aren't registered anywhere but have been in use for quite a while *(decades I believe)*.

Note(s):
* Missing defaulted Linux installation entries with `.pyx`, `.wsgi` at `text/x-python` in Debian source tree.
* Missing defaulted Linux installation entries with `.py3`, `.py3x` at `text/x-python3` in Debian source tree ... probably due to transitional nature of Python 2 vs Python 3. I see this script language a lot of scripting circles.
* https://wiki.debian.org/MimeTypesSupport
* Can't find on IANA, Apache, or nginx refs you posted. It is a scripting language and that's what is in my File Associations in KDE *(alternatively /etc/mime.types in the tree with Ubuntu/Kubuntu)*.
* https://www.python.org ... search [seems to yield only that they support MIME types](https://www.python.org/search/?q=mime+type&submit=) and mostly not their own. Note this seems to be a front end for a mailing list going back a while in history.
* https://www.python.org/dev/peps/pep-0441/ Shows zipped `.py` file extension of `.pyz` ... again no MIME type mentioned.

dougwilson:
Hi @syscaday sorry for the confusion. It sounds like you are trying to get a media type added that just isn't registered or is not consistent. The process to do this is not through a PR, instead the instructions you can find at the bottom of our readme: https://github.com/jshttp/mime-db#adding-custom-media-types

src/custom-types.json
dougwilson: Hi @Martii thanks for this. That debian source is just an aggregate list, the quality of which is unknown. If we can better understand how they sourced those types and their process to maintain it, we can consider adding it as a source here, though though the custom type process, that would be considered a third-party reference.

Unfortunately some programs/projects do not feel they need to make a media type, so they never state one or register one. Then folks build programs around media types and just make up ones for file extensions they care about but have no declared one, so you end up having several variations of a media type for a file extension, but this typically stems from an internal need from some program to have a media type. The reality is that file extensions can exist without media types and vice versa. A file extension without some standard type is not useful for external interoperability. If the Python project thinks their files should make a media type... perhaps they should just state one? I believe that Python is a very active language community and so the absence of one from the project may be that they don't want there to be one, I'm not sure, but I think we have the ability to actively ask them what is the media type for their files.

Martii: Rehi @dougwilson 

Yah... I read a little further down on the .pyz at https://www.python.org/dev/peps/pep-0441/#registering-pyz-as-a-media-type and that "thing" *(PEP)* mentions:

> It doesn't seem reasonable to register .pyz without .py,

Thanks for the clarifications.

src/custom-types.json
syscaday: text/x-python-script really cannot get the reference, but not sure why server was returning this. we manually handle this.

src/custom-types.json
syscaday: i had change my source URL kindly review again

PR #241, opened 2021-07-01
Add .py extension

syscaday:
possible to look on this again

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -837,6 +837,12 @@
   "text/x-scss": {
     "extensions": ["scss"]
   },
+  "text/x-python": {
+    "extensions": ["py", "pyx", "wsgi"],
+    "sources": [
+      "https://reposcope.com/mimetype/text/x-python"
+    ]
+  },
   "text/x-suse-ymp": {
     "compressible": true,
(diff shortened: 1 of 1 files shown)


src/custom-types.json
```diff
@@ -837,6 +837,12 @@
   "text/x-scss": {
     "extensions": ["scss"]
   },
+  "text/x-python": {
+    "extensions": ["py", "pyx", "wsgi"],
+    "sources": [
+      "https://reposcope.com/mimetype/text/x-python"
```
dougwilson, line 843: At best, this is not a primary source (rather, a secondary source), but I think it is actually the explicitly disallowed "type aggregating sites" as it seems I can change the URL to other mime types to see information about those as well (see https://reposcope.com/mimetype/text/plain). The link should be to how that site sourced their data on why those extensions are that mime type. The ideal primary source would be a link on the python project's own site.

dougwilson (changes requested):

src/custom-types.json
syscaday: thanks for yr info, that mean i only can do in manually look up in my project I think

PR #244, opened 2021-09-06
Added deprecated iWorks content types with extensions

rhodgkins:
They are deprecated but I've found places still return them (Google Drive API being one).

Reference from (under _Additional information_):
https://www.iana.org/assignments/media-types/application/vnd.apple.keynote
https://www.iana.org/assignments/media-types/application/vnd.apple.numbers
https://www.iana.org/assignments/media-types/application/vnd.apple.pages

I've also updated the _unreleased_ section of the `HISTORY.md` - hope that's ok!

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,9 @@
 unreleased
 ==========
 
+  * Add `application/x-iwork-keynote-sffkey`
+  * Add `application/x-iwork-numbers-sffnumbers`
+  * Add `application/x-iwork-pages-sffpages`
   * Add new upstream MIME types
 
 1.49.0 / 2021-07-26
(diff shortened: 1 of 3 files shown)


dougwilson (approved):
Thank you, very much appreciated!

rhodgkins:
Thanks @dougwilson - was hoping this would be picked up by `mime-types` on an upgrade, but just noticed that the [dependency](https://github.com/jshttp/mime-types/blob/7398cf53b061adb1a82f95366a04b7b4b998c468/package.json#L17) is a fixed version - how come?

Is it possible to get a new release of `mime-types` with this version of `mime-db` please?

PR #248, opened 2021-10-07
Indicate that .DDS textures are compressible

donmccurdy:
I'm working with some large .DDS image-based lighting (IBL) textures, and found that my static server isn't compressing them (via [`compressible`](https://www.npmjs.com/package/compressible)). DDS is a container format for various GPU texture compression formats and I can't be sure that all variations benefit from compression, but the samples I have go from 8MB → 1.2MB with gzip. IBL files can get pretty large, so I think this is likely to be preferable for most WebGL/WebGPU applications.

/cc @the-simian (author of #170)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@ unreleased
 
   * Add new upstream MIME types
   * Mark `image/vnd.microsoft.icon` as compressible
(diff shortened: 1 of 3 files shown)


dougwilson (approved):
Thanks

PR #249, opened 2021-10-26
adds x-zip-compressed

jlandry:
This is an archaic zip compression application type that works with the .zip extension

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -451,10 +451,7 @@
   "application/x-zip-compressed": {
     "compressible": false,
     "extensions": ["zip"],
-    "notes": "winZip file format",
-    "sources": [
-      "https://bugs.python.org/issue41738"
-    ]
+    "notes": "winZip file format"
   },
   "application/xml": {
     "compressible": true,
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -453,7 +453,7 @@
     "extensions": ["zip"],
     "notes": "winZip file format",
     "sources": [
-      "https://mimetype.io/application/x-zip-compressed"
+      "https://bugs.python.org/issue41738"
     ]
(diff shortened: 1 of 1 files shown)


nihey:
@dougwilson Can we have this merged now that the sources have been removed? I've experienced this same problem with a product that I'm working on. It seems that Chrome on windows always tells that `.zip` files are `application/x-zip-compressed` mimetype files.

I've setup a Simple JSFiddle in case anyone wants to test it: http://jsfiddle.net/bn32mzw1/6/

The file mimetype will be shown below the input after selecting it on this demo.

A file `compositions.zip` on Linux:

Screenshot from 2021-10-29 16-50-43

The exact same file on windows:

photo1635537012

Screenshot from 2021-10-29 16-52-02

dougwilson:
Hi @nihey this module is meant to be an aggregation of the three sources listed a the top of the readme. There are so many extensions and types, it is difficult to keep track of, I understand. This module is just a helper to aggregate those three sources, not a registry service for new mime types; a registry service that feeds this module is listed in the README.

As a courtesy, we accept entries directly here if there is a concrete source for them. I tried out your Fiddle on Windows 10 and I got `application/zip`. Ultimately, the types from Windows is just based on what is in the registry that any application you install can change as it would like.

If there is not actual documented relation of this type, or it doesn't appear in one of the three databases here, then it just won't appear in this module. Whatever application system you are using that utilizes this module should provide a method to add additional custom type mappings, as you will always end up encountering mappings that won't show up here.

@jlandry I see your notes say "WinZip file format" -- can you provide a link to some WinZip documentation that says .zip file extension == application/x-zip-compressed ?

dougwilson:
Closing as this is a draft pr and no further responses.

PR #250, opened 2021-11-04
Ensure image/vnd.microsoft.icon is marked as compressible

ijjk:
Hi, this updates `image/vnd.microsoft.icon` to be marked as `compressible` as `image/x-icon` is already marked as compressible and according to the doc [from IANA](https://www.iana.org/assignments/media-types/image/vnd.microsoft.icon) they are the same although `image/x-icon` seems to be used more commonly. The source used for `image/x-icon` also mentions the `image/vnd.microsoft.icon` type being used [here](http://en.wikipedia.org/wiki/ICO_(file_format)).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@ unreleased
 ==========
 
   * Add new upstream MIME types
+  * Mark `image/vnd.microsoft.icon` as compressible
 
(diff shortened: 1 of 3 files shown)


dougwilson (approved):
Thanks!

PR #252, opened 2022-01-24
Add Typescript Definitions for all mime types

JustusFluegel:
It would be quite nice to have all mime types as a typescript literal type as well. See changelog.


dougwilson:
Thank you for this, but it doesn't look like this is definitions for how to use this module, but rather definitions for a generic type derived from this module. I would suggest publishing this as it's own module for the usage you are trying to use it for.

JustusFluegel:
kk, will do

JustusFluegel:
If it's welcome to add definitions I'll add definitions for the usage of the Module tomorrow, as it's 1:20 in Germany and I am going to sleep for now.

PR #254, opened 2022-01-24
Add types for mime-db

JustusFluegel:
Add types for the default export of the module.

Signed-off-by: Justus Fluegel <justusfluegel@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,18 @@ https://cdn.jsdelivr.net/gh/jshttp/mime-db@master/db.json
 ## Usage
 
 ```js
-var db = require('mime-db')
+var db = require("mime-db");
 
 // grab data on .js files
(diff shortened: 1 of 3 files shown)


dougwilson:
Thank you! It seems there is already a `@types/mime-db` that defines types for this module. When I pull these in here, folks will get this new definition instead. I'm not a big user of typescript, so not sure of the answer if you can help: will these types break projects that are using the existing types?

zsakowitz:
I've had some experience working with TypeScript and I think that directly pulling the typings from `@types/mime-db` into this package won't affect existing users.

dougwilson:
Hi @zsakowitz I think you misread what I wrote / didn't check the PR. This PR has different typings from what everyone is already using. I'm going to close this because I never heard back from OP.

PR #255, opened 2022-01-27
Added the `wav` extension to `audio/vnd.wave`

aarongeorge:
Resolves issues with https://github.com/oakserver/media_types/pull/12

### DIFF
--- a/db.json
+++ b/db.json
@@ -6850,8 +6850,7 @@
     "source": "iana"
   },
   "audio/vnd.wave": {
-    "compressible": false,
-    "extensions": ["wav"]
+    "compressible": false
   },
   "audio/vorbis": {
     "source": "iana",
--- a/db.json
+++ b/db.json
@@ -6850,7 +6850,8 @@
     "source": "iana"
   },
   "audio/vnd.wave": {
-    "compressible": false
+    "compressible": false,
+    "extensions": ["wav"]
   },
   "audio/vorbis": {
     "source": "iana",
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -504,7 +504,8 @@
     "compressible": false,
     "extensions": ["wav"],
     "sources": [
-      "https://www.loc.gov/preservation/digital/formats/fdd/fdd000001.shtml"
+      "https://www.loc.gov/preservation/digital/formats/fdd/fdd000001.shtml",
+      "https://bugs.webkit.org/show_bug.cgi?id=173635"
     ]
   },
   "audio/vorbis": {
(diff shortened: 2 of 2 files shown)


dougwilson (changes requested):
Please include an array of URLs as the `sources` of where the MIME type and the associated extensions are sourced from. This needs to be a primary source; links to type aggregating sites and Wikipedia are not acceptable.

aarongeorge:
@dougwilson Sorry, I have added a source

dougwilson:
Thanks. Please add a primary source. That seems like a secondary source at best.

aarongeorge:
@dougwilson I've linked the webkit bug. Is that primary enough?

dougwilson:
No, i don't think so. Can you explain how that would be a primary source?

aarongeorge:
Fair enough. Is the ietf.org source primary?

I'm not trying to be difficult. I'm trying to resolve an issue in a downstream project that relies on your DB.

dougwilson:
Thanks. I don't see anywhere in that document that is says the file extension you are adding is associated to this mime type, but to be fair it is a large doc. Can you point me to that piece of information?

dougwilson:
Reading that document it seems that the ;codec=YYY is required, but there are a lot to choose from. I guess that is why it does not declare any file extension for this, as one would not have a way to know what the code value would be from such a simple mapping.

aarongeorge:
I'm not sure what else to say really. Multiple apps use the raw `audio/vnd.wave` as a mime-type and people seem to not wish to use anything other than your DB as a resource and so it's breaking many tools that require uploading a .wav file.

Any ideas?

aarongeorge:
As far as the web and mime-types are concerned it seems like they treat the properties as optional. https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

dougwilson:
Since this module is the aggregate of the threw data sources listed at the top of the readme, perhaps you should get this added to one of them and it'll just pull in here?

As a curtosy, we do accept entries directly, but only with definite sources, as we are not here to argue about these things. This was just made to aggregate the three sources for our usage, made available to whoever chooses to use it under MIT license.

As far as others using this module, I cannot speak for them. I have no contact with anyone and if they want to use it, they are free to or not. What this module does is written in the readme (aggregates from the three daya sources).

Not all media types have file extensions (just look at the db). I'm not sure who you are and what authority you have to declare that this media type does indee have a file extension when the documentation you provides about the media type doesn't have one.

Like I said, the goal of this module is not to be a registry, arbitrator or anything similar, just was a little project to aggregate from the three sources and make a json file. Perhaps plead you case to those sources.

PR #258, opened 2022-03-21
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -25,7 +25,7 @@ Object.keys(mime).forEach(function (suffix) {
   var s = mime[suffix]
 
   Object.keys(db).forEach(function (type) {
-    if (type.substr(0 - suffix.length) !== suffix) {
+    if (type.slice(-suffix.length) !== suffix) {
       return
     }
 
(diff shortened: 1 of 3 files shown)


scripts/build.js
```diff
@@ -25,7 +25,7 @@ Object.keys(mime).forEach(function (suffix) {
   var s = mime[suffix]
 
   Object.keys(db).forEach(function (type) {
-    if (type.substr(0 - suffix.length) !== suffix) {
+    if (type.slice(0 - suffix.length) !== suffix) {
```
zsakowitz, line 28: I think you can use `-suffix.length` instead of `0 - suffix.length` on this line.

CommanderRoot: You totally can. I have updated the PR

dougwilson (approved):

PR #262, opened 2022-05-07
Correct JavaScript MIME types + extensions per RFC 9239

mathiasbynens:
This patch updates the MIME type configuration per RFC 9239.
https://www.rfc-editor.org/rfc/rfc9239

First, the recommended MIME type is now `text/javascript`:

> The most widely supported media type in use is `text/javascript`; all
> others are considered historical and obsolete aliases of `text/javascript`.

Second, the `.mjs` extension is now explicitly registered:

> The `.mjs` file extension signals that the file represents a JavaScript
> module. Execution environments that rely on file extensions to
> determine how to process inputs parse `.mjs` files using the Module
> grammar of [ECMA-262].

(mime-db already included the `.mjs` extension prior to this patch, so no
change needed here.)

IANA template: https://www.iana.org/assignments/media-types/text/javascript

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,7 @@ unreleased
   * Add extensions from IANA for more MIME types
   * Add new upstream MIME types
   * Fix extensions for `text/markdown` to match IANA
+  * Remove extension `.mjs` from `application/javascript`
   * Remove obsolete MIME types from IANA data
 
 1.52.0 / 2022-02-21
--- a/db.json
+++ b/db.json
@@ -735,7 +735,7 @@
     "source": "apache",
     "charset": "UTF-8",
     "compressible": true,
-    "extensions": ["js","mjs"]
+    "extensions": ["js"]
   },
   "application/jf2feed+json": {
     "source": "iana",
@@ -7859,6 +7859,7 @@
   },
   "text/javascript": {
     "source": "iana",
+    "charset": "UTF-8",
     "compressible": true,
     "extensions": ["js","mjs"]
   },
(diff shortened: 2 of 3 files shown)


mathiasbynens:
Related PRs:

- [apache/httpd PR](https://github.com/apache/httpd/pull/318)
- [nginx/nginx PR](https://github.com/nginx/nginx/pull/77)

dougwilson (changes requested):
Hello, and thank you for your pull request. Unfortunately this cannot be accepted as-is, as the `db.json` file is a generated file and changes made to it directly will just be overwritten when `npm run build` is run. Please see https://github.com/jshttp/mime-db#contributing for how to make contributions to the database.

mathiasbynens:
@dougwilson Thanks. It looks like it’s been a while since the scripts have been run, since there are a bunch of changes unrelated to this new RFC. I also noticed that some tweaks to the scripts are required in order to even pick up the new RFC.

Here’s what I think we should do, in order:

- [ ] Update IANA types by running scripts/fetch-iana: #264
- [ ] Allow scripts/fetch-iana to follow redirects: #263
- [ ] Remove the hardcoded and now unnecessary entries from `custom-types.json`: TODO

mathiasbynens:
Thanks @dougwilson! It looks like https://github.com/jshttp/mime-db/commit/c8e4e30da4571d55003f7d582ce4cd7080ab26b0 picks up the upstream `text/javascript` changes 🎉 What’s still missing is the `extensions` property set to `[".js", ".mjs"]` (even though both extensions are explicitly listed in the IANA registrations). I’ve updated this PR to fix that by adding it to `custom-types.json`.

dougwilson (approved):

Issue #194, opened 2020-05-01
JavaScript type should be `text/javascript`


broofa:
Apparently `text/javascript` is the new hotness now...?  https://datatracker.ietf.org/doc/draft-ietf-dispatch-javascript-mjs/

I first noticed this on MDN's [Common MIME Types](https://wiki.developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) page, which now shows `text/javascript`.

@mathiasbynens, @domenic: Can you provide insight into what's going on at the standards body level?  When do you expect to see the IANA registry updated?  What about mime.types for Apache or NGINX?  (mime-db pulls from all three of these places.)

mathiasbynens:
Thanks for the ping!

This is not really "new"; the HTML Standard has included [a willful violation of RFC 4329](https://html.spec.whatwg.org/multipage/#dependencies:willful-violation) for a long time:

> The term "JavaScript" is used to refer to ECMA-262, rather than the official term ECMAScript, since the term JavaScript is more widely known. Similarly, the MIME type used to refer to JavaScript in this specification is `text/javascript`, since that is the most commonly used type, despite it being an officially obsoleted type according to RFC 4329. [RFC4329]

[And](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages):

> Servers should use `text/javascript` for JavaScript resources. Servers should not use other JavaScript MIME types for JavaScript resources, and must not use non-JavaScript MIME types.

RFC 4329 does not match reality. The HTML Standard does (in this particular case), and https://datatracker.ietf.org/doc/draft-ietf-dispatch-javascript-mjs/ is in the final stages of being standardized so that RFC 4329 is officially deprecated at the IETF/IANA level.

There's a bit of a chicken-egg problem in that server software often has a policy of only considering IETF/IANA even when those standards are clearly out of touch with reality, like in this case. This leads to ridiculous situations like Python's built-in `SimpleHTTPServer` adding support for `.mjs` with the wrong `application/javascript` MIME type. 🤷‍♂️

I hope you consider changing changing `mime-db`'s without waiting for https://datatracker.ietf.org/doc/draft-ietf-dispatch-javascript-mjs/ to be finalized.

dougwilson:
Ultimately this module is not a MIME registry service and not in the business of handling such disputes. It is an aggregation of the sources in the readme, which is where these types of changes need to be made and this module will passively update the records to reflect changes in the source registries.

[...]

PR #263, opened 2022-05-08
Allow redirects in fetch-iana script

mathiasbynens:
Certain URLs of the form `https://tools.ietf.org/rfc/${RFC_ID}.txt` redirect multiple times before arriving to the canonical resource. Example:

    $ curl -sI 'https://tools.ietf.org/rfc/rfc9239.txt' | grep Location
    Location: https://rfc-editor.org/rfc/rfc9239.txt

 $ curl -sI 'https://rfc-editor.org/rfc/rfc9239.txt' | grep Location
    Location: https://www.rfc-editor.org/rfc/rfc9239.txt

    $ curl -sI 'https://www.rfc-editor.org/rfc/rfc9239.txt' | grep HTTP
    HTTP/1.1 200 OK

`cogent`’s implicit default is to only follow 1 redirect. This patch explicitly configures `cogent` to follow up to two redirects.

This is required to properly fetch RFC 9239 and update the JavaScript MIME type entries.

Issue: #194, #262


mathiasbynens:
#264 should be merged first. Then I can update this patch to include only the `src/iana-types.json` diff caused by these changes.

dougwilson:
Hello, and thank you for this PR. I'm not seeing any different behavior before or after your PR, and the default redirects for the `cogent` module we used in 3, which is larger than you are setting here. Is there a particular reason this should only allow up to 2 redirects instead of the default 3 redirects? https://github.com/cojs/cogent/blob/master/lib/index.js#L20

mathiasbynens:
Ah, as stated in the commit message, my assumption was that cogent defaults to following only 1 redirect. I got that from [this README](https://github.com/cojs/cogent#var-response--yield-requesturl-options), which states:

> `redirects` - resolve redirects, default `1`

If the true default is indeed `3` then we don’t actually need this. Feel free to close this.

mathiasbynens:
I assumed that this was the cause of the newer entries w.r.t. RFC 9239 not showing up. I guess that was wrong :)

Issue #274, opened 2022-10-03
YAML is application/yaml


ioggstream:
## I expect

- YAML media type registration is going to file application/yaml See https://datatracker.ietf.org/doc/draft-ietf-httpapi-yaml-mediatypes/

## Instead

- it is published as text/yaml

dougwilson:
Hello, and thank you for providing that. It looks like that is still a draft, and so no guarantee that is going to be finalized. Once the registration is published, this module will pull it in automatically from IANA.

ioggstream:
@dougwilson OK, thanks!

dougwilson:
No problem! We have seen a few times over the years where some really ingrained type that was never registered had a draft RFC to change it to something else, but ultimately was abandoned due to friction to change it in reality. So we'll see if this actually gets published fully (and then it'll just be auto pulled in anyway by our feed to IANA).

PR #275, opened 2022-10-03
Fix: #274. Use application/yaml.

ioggstream:
## This PR

- text/yaml -> application/yaml according to https://www.ietf.org/archive/id/draft-ietf-httpapi-yaml-mediatypes-03.html


dougwilson:
Hello, and thank you for your PR. Please see my response on your original issue; we'll need the registration to leave the draft state before updating here.

PR #281, opened 2023-01-25
add jfif to image/jpeg

chooklii:
"jfif" is missing as a file-type for image/jpeg 
https://www.w3.org/Graphics/JPEG/


chooklii:
Thanks for the Info! - After additional research I found out that jfiff is only a "defacto" standard and no real standard for mime-types. (RFC2035 and no valid RFC). So this PR and the addition of jfiff as a type is unnecessary

Issue #176, opened 2019-10-29
IPython notebooks extension support


dmvieira:
We have no sources with this standard for ipynb extension: https://jupyter.readthedocs.io/en/latest/reference/mimetype.html

dougwilson:
I don't see anything about a `.ipynb` file extension on that page?

dmvieira:
I see that it's alread in [shared mime info](https://freedesktop.org/wiki/Software/shared-mime-info/) in this [line](https://gitlab.freedesktop.org/xdg/shared-mime-info/blob/master/freedesktop.org.xml.in#L2520). What do you think about include this source in mime-db?

dougwilson:
This module only pulls from the three sources listed at the top of the readme. To pull it in, the best method is relly to get it into the IANA database.

dougwilson:
As a stop gap measure, types can be added directly to this module, but you would need to find a primary source so I can vet it. Sites that are just aggregate mime types are a secondary source. So if you can find out where that site sourced the data from for this mapping that should work.

flying-sheep:
Well, that’s a mirror of freedesktop’s “[shared mime info](https://freedesktop.org/wiki/Software/shared-mime-info/)” list. freedesktop is an initiative for interoperability in *NIXoids, to which end they define standards. They aren’t an official standards body, but followed by all bigger desktop environments on Linux like GNOME and KDE Plasma.

I consider a DB compiled by them second to IANA, and much more official than any specific server software’s collection. So if you aggregate from nginx and Apache, you should definitely also aggregate from freedesktop’s shared-mime-info.

dougwilson:
If we were to add that as a source, where is the shared mime info list hosted for us to pull from it directly? Do they have any policy against automated pulling of their data?

flying-sheep:
On the linked page are “[Sources](https://gitlab.freedesktop.org/xdg/shared-mime-info/)” and “[Releases](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/releases)”, both linking to their GitLab instance. Linux distros [like e.g. Arch pull](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/shared-mime-info) from there.

dougwilson:
[...]

PR #282, opened 2023-01-25
Add mime-type for Jupyter notebooks

julien-c:


src/custom-types.json
```diff
@@ -295,6 +295,13 @@
       "http://php.net/manual/en/faq.installation.php"
     ]
   },
+  "application/x-ipynb+json": {
+    "compressible": true,
+    "extensions": ["ipynb"],
```
dougwilson, line 300: Thank you for your contribution! I checked the reference below, and didn't find anything about this being the file extension for the `application/x-ipynb+json` mime type. But I could be missing something, of course. If you need to add another reference for this, please do so. If it is already on that page, but I am just missing it, please highlight where this information is location on the source linked below. Thank you!

dougwilson (changes requested):

src/custom-types.json
julien-c: Switched to a better source, and also noting that this would close the previously discussed #176

dougwilson: Sweet, thank you! Please forgive my misunderstanding still, but the issue comment you linked to says:

> x- or vnd are fine - they're both valid for these "custom" (non-IANA declared). I primarily care about ensuring consistency and documented specs. 😄
> 
> application/x-ipynb+json was declared sometime ago, we can keep on with it.

I guess I'm missing where in that comment it talks about the file extension??

julien-c: Oh, I see! sorry for not understanding the comment about the file extension.

Well, it's actually surprisingly tricky to find a reference for the file extension being .ipynb, but maybe [this link](https://docs.jupyter.org/en/latest/running.html#:~:text=The%20following%20code%20should%20open%20the%20given%20notebook%20in%20the%20currently%20running%20notebook%20server%2C%20starting%20one%20if%20necessary.)?

dougwilson: Hmm. The issue is that we have a strong rule on needing a specific reference that clearly provide the information like where it is clear that the specific file extension == that specific mime type. I'm not familiar with Jupyer and there is just not enough context there to link the two together definitively. It has been too frequent that we would accept data only for folks to come argue with us it was wrong or should be changed, over and over. This module is not intended to be a registry, that is what IANA is; it's main purpose is to aggregate the three sources listed at the top of the README in to a JSON file for Node.js users like the Express.js project (the spawn of it).

We of course still accept custom entries, but need good sources from first parties regarding the data before landing so we're not here in a situation of being a registry. I hope that makes sense.

julien-c: Got it. 

If I ask the Jupyter maintainers if they'd state somewhere in their doc the following language:

> Jupyter notebooks use file extension `.ipynb` and their intended mime-type is `application/x-ipynb+json`

Would that work?

dougwilson: Absolutely 💯 even just as a comment on their Github would be enough. We have a few like that currently.

wesleytodd (approved):
Looks like the last update is from here: https://github.com/jupyter/jupyter/issues/68

I am personally a bit more comfortable with this inclusion reading the [attached sources](https://bugs.freedesktop.org/show_bug.cgi?id=94878). Particularly so because I found a ton of references across multiple sources confirming this is *widely used* as the correct extension.

I am going to leave this approval and look for any new blocking reviews, if I don't see any then I will merge this for the next release.

bjohansebas (approved):

julien-c:
Yay! 🔥

PR #285, opened 2023-02-15
fix: Mark text/x-component as compressible

shuding:
According to https://en.wikipedia.org/wiki/HTML_Components and other resources, `text/x-component` (`.htc` file) is plain text in XML and also very similar to HTML, and it is compressible.

### DIFF
--- a/db.json
+++ b/db.json
@@ -8305,6 +8305,7 @@
   },
   "text/x-component": {
     "source": "nginx",
+    "compressible": true,
     "extensions": ["htc"]
   },
   "text/x-fortran": {
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,6 +11,7 @@ unreleased
   * Fix extensions for `text/markdown` to match IANA
   * Remove extension `.mjs` from `application/javascript`
   * Remove obsolete MIME types from IANA data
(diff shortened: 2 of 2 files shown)


dougwilson (changes requested):
Hello, and thank you for your contribution! You cannot modify the `db.json` file directly, as it is a generated file and changes you make will be overwritten. Please see contribution instructions at https://github.com/jshttp/mime-db#contributing . When you update the custom file, please include the source of your information, which cannot be Wikipedia (for example, you can put in the source that Wikipedia cited for that information).

shuding:
Hi @dougwilson, I just updated the PR to follow the contributing guidelines with sources added.

db.json
```diff
@@ -8305,6 +8305,7 @@
   },
   "text/x-component": {
     "source": "nginx",
+    "compressible": true,
```
wesleytodd, line 8308: This must be removed.

wesleytodd (changes requested):

db.json
bjohansebas:
```suggestion
```

bjohansebas (approved):
LGTM

PR #287, opened 2023-02-16
Update mime type for DCM format

darshitmaniar:
Kindly update this block

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -503,9 +503,8 @@
   },
   "application/dicom": {
     "source": [
-        "https://en.wikipedia.org/wiki/DICOM#"
+        "https://en.wikipedia.org/wiki/DICOM"
       ],
-    "compressible": true,
     "extensions": ["dcm"]
   },
   "audio/3gpp": {
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -501,6 +501,13 @@
   "application/zip": {
     "compressible": false
   },
+  "application/dicom": {
+    "source": [
+        "https://en.wikipedia.org/wiki/DICOM#"
+      ],
+    "compressible": true,
+    "extensions": ["dcm"]
+  },
   "audio/3gpp": {
     "compressible": false,
     "extensions": ["3gpp"]
--- a/db.json
+++ b/db.json
@@ -428,9 +428,7 @@
     "compressible": true
   },
   "application/dicom": {
-    "source": "iana",
-    "compressible": true,
-    "extensions": ["dcm"]
+    "source": "iana"
   },
   "application/dicom+json": {
     "source": "iana",
--- a/db.json
+++ b/db.json
@@ -428,7 +428,9 @@
     "compressible": true
   },
   "application/dicom": {
-    "source": "iana"
+    "source": "iana",
+    "compressible": true,
+    "extensions": ["dcm"]
   },
   "application/dicom+json": {
     "source": "iana",


dougwilson (changes requested):
Hello, and thank you for your contribution! You cannot modify the `db.json` file directly, as it is a generated file and changes you make will be overwritten. Please see contribution instructions at https://github.com/jshttp/mime-db#contributing . When you update the custom file, please include the source of your information, which should clearly state the extension is for this mime type and not be a prohibited site. Thank you!

darshitmaniar:
Thanks for the reply, update the `custom-types.json` file as per instructions.

src/custom-types.json
```diff
@@ -501,6 +501,13 @@
   "application/zip": {
     "compressible": false
   },
+  "application/dicom": {
+    "source": [
+        "https://en.wikipedia.org/wiki/DICOM#"
```
dougwilson, line 506: Hello, and thank you. The contribution guide should be clear that links to Wikipedia are not acceptable. Please link to the primary source, i.e. where that Wikipedia sourced the information for this.

dougwilson (changes requested):

src/custom-types.json
dougwilson: I forgot to add that I see you also put `compressible: true`, so please make sure there is also at least one source included for how the determination of compressible was determined.

src/custom-types.json
```diff
@@ -501,6 +501,12 @@
   "application/zip": {
     "compressible": false
   },
+  "application/dicom": {
+    "source": [
+        "https://en.wikipedia.org/wiki/DICOM"
```
dougwilson, line 506: This needs to be a link to a primary source; Wikipedia is even specifically called out as not acceptable in the README: https://github.com/jshttp/mime-db#direct-inclusion

dougwilson (changes requested):

src/custom-types.json
DimitriPapadopoulos: See instead:
* https://www.dicomstandard.org/concepts
  > DICOM files typically have a .dcm file extension if they are not part of a DICOM media (which requires them to be without extension). 
* [DICOM PS3.12 2023c - Media Formats and Physical Media for Media Interchange](https://www.dicomstandard.org/standards/view/media-formats-and-physical-media-for-media-interchange)
  > * "name" is constructed from the last DICOM File ID component (that means the "file name" without "path" information) and the extension ".dcm" (except for the DICOMDIR).
  > 
  > For example: "I00023.dcm"
* [RFC 3240](https://www.rfc-editor.org/info/rfc3240)
  > 2. File extension(s): ".dcm" is recommended for files saved to disk (other than DICOMDIR)

DimitriPapadopoulos: Additionally, `"application/dicom"` is already present in [`src/iana-types.json`](https://github.com/jshttp/mime-db/blob/master/src/iana-types.json):
https://github.com/jshttp/mime-db/blob/a76e5a824c228e2e58363c9404e42a54ee1d142f/src/iana-types.json#L607-L612

You should probably add `"extensions": ["dcm"]` at the above location, like #295.

DimitriPapadopoulos:
@darshitmaniar @dougwilson I have the commits in this PR and modified the primary source in #362. Feel free to integrate my commit here or to switch to PR #362.

UlisesGascon:
I will close this as the discussion will take place now in https://github.com/jshttp/mime-db/pull/362.

Thanks @darshitmaniar for creating the new PR :heart:

Issue #283, opened 2023-02-07
Add image/jpeg for .jfif file extension


kaushals-deck:
Hello,
Please add the jfif file extension/type also to image/jpeg. As per Wiki
https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format

PR #291, opened 2023-04-28
Add jpe,jif,jfif as image/jpeg mime type

ianchanning:
jfif is the only one that I have actual files for, but for completeness based on https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#jpeg_joint_photographic_experts_group_image I included jpe and jif.

Fix for #283

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -634,7 +634,11 @@
     ]
   },
   "image/jpeg": {
-    "compressible": false
+    "compressible": false,
+    "extensions": ["jpe", "jif", "jfif"],
+    "sources": [
+      "https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#jpeg_joint_photographic_experts_group_image"
+    ]
   },
   "image/jpm": {
     "compressible": false,


src/custom-types.json
```diff
@@ -634,7 +634,11 @@
     ]
   },
   "image/jpeg": {
-    "compressible": false
+    "compressible": false,
+    "extensions": ["jpe", "jif", "jfif"],
+    "sources": [
+      "https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#jpeg_joint_photographic_experts_group_image"
```
dougwilson, line 640: We need the primary source to be linked fom where these are defined. The MDN site is a secondary source.

dougwilson (changes requested):

src/custom-types.json
ianchanning: Its frustrating because it's not mentioned on the IANA site, but it is an ISO standard.

This is from the official JPEG site <https://jpeg.org/jpeg/workplan.html>:

> ITU-T T.871 (2011) — ISO/IEC 10918-5:2013:
> JPEG File Interchange Format (JFIF)
> ([ITU page](https://www.itu.int/ITU-T/recommendations/rec.aspx?id=11319), ITU edition 1, [ISO page](https://www.iso.org/standard/54989.html?browse=tc), ISO edition 1, in stage [90.93](https://www.iso.org/home/developing-standards/resources/stage-codes.html#90.93))

Is the [ISO page](https://www.iso.org/standard/54989.html?browse=tc) good enough?

There's also this via the [JFIF Wikipedia page](https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format): <https://www.w3.org/Graphics/JPEG/jfif3.pdf>

dougwilson: Gotcha. I wonder where did the MDN page get it then? Maybe I am confused, but your quotes does not say that JFIF is a file extension, just the abbreviation of JPEG File Interchange Format.

ianchanning: It clearly is a file extension for jpeg files same as this Adobe page talks about it https://www.adobe.com/creativecloud/file-types/image/raster/jpeg-file.html - but it doesn't seem like it was standardised anywhere.

dougwilson: Gotcha. Unfortunately this just isn't the purpose of this module, to be a registry -- it is built to be a JSON version of the three sources listed at the top of the module's README. We can accept other ones that are not in there, but it needs to be very clear standard or like something is broken with our scraping of those three sources. I suggest the best way forward without some kind of confirmation from JPEG itself that someone get it registered in IANA. This module is slowing eliminating all entries not included in those three sources, and this seems contrary to the module's direction.

dougwilson:
Sorry, did not mean to close the PR.

src/custom-types.json
dougwilson: Ideally for file extensions, we are looking for a first-party source which clearly states that "file extension X represents mime type Y"

dougwilson: Overall, I would like to ideally see if we can get these added to one of our upstream sources or fine some page to link to. I actually emailed JPEG to ask them, since it wasn't clear if you did that or not already. I'll see what they say.

ianchanning: As I see it there is no primary source for 'jpg' being an official extension. IANA doesn't include it. You include it because Apache does and they also include 'jpe' - but with no reason given.

```
image/jpeg					jpeg jpg jpe
```

JPEG don't specify any file extensions, Windows just decided to use jpg to keep it short and everyone else jpeg.

So you don't think you can find a primary source for 'jpg'.

So I doubt we'll get much out of JPEG - I don't think they care about the extension.

ianchanning: Actually now that I've seen that `jpe` is included in the Apache list I can remove it from here.

I'm happy if you want to close this - it makes sense to have a clear definition for what is or isn't included here.

DeniKungurov:
please tell me what is the status of the task

mramosbhz:
So, what do we need to do to add JFIF as an acceptable extension for image/jpeg?

ianchanning:
> So, what do we need to do to add JFIF as an acceptable extension for image/jpeg?

Something from IANA, Apache or NGINX that specifies JFIF as a mimetype. It seems that nothing from either JPEG group themselves or IANA specify it. If you need to have jfif specified as a mimetype you will have to look for other libraries that do things like sniffing the binary contents of the files.

ianchanning:
... or request a new mime type: https://www.iana.org/form/media-types - I don't know the answers to all their questions.

mramosbhz:
> > So, what do we need to do to add JFIF as an acceptable extension for image/jpeg?
> 
> Something from IANA, Apache or NGINX that specifies JFIF as a mimetype. It seems that nothing from either JPEG group themselves or IANA specify it. If you need to have jfif specified as a mimetype you will have to look for other libraries that do things like sniffing the binary contents of the files.

Oh! But JFIF is not a mime type. Is an extension for `image/jpeg` mime type. 🤷‍♂️
We're using a lib called mime-types (from `jshttp` too) that try to find the file mime-type based on the file extension. This lib uses mime-db as a resource. So it fails!
I think I'll have to wrap the call of the lib to manage only this file extension 😞

Issue #292, opened 2023-04-28
Missing extensions list for model/step


JohnSimumatik:
Type `model/step` in db.json is missing an extensions list. 
Should be `.p21`, `.stp`, `.step`, `.stpnc`, `.210` according to https://www.iana.org/assignments/media-types/model/step

dougwilson:
Thanks! We don't (yet) support pulling in a large list from iana yet, would you be willing to make a pr that adds these? https://github.com/jshttp/mime-db#direct-inclusion

JohnSimumatik:
@dougwilson opened a PR #293

PR #293, opened 2023-05-02
Added extensions list for model/step

JohnSimumatik:
model/step was missing the extensions list. I added the extensions listed here: https://www.iana.org/assignments/media-types/model/step

Fix for #292


dougwilson (changes requested):
Hello and thank you! But you cannot modify `db.json` as that is a generated file and your changes will get overridden. Please update your PR to follow the instructions at https://github.com/jshttp/mime-db#direct-inclusion

PR #295, opened 2023-06-15
Update iana-types.json

pzsette:
Add 'dcm' extension to application/dicom mime. 'dcm' is the recommended extension for this multimedia type as stated at point 2 of additional information in this iana.org page https://www.iana.org/assignments/media-types/application/dicom.

### DIFF
--- a/src/iana-types.json
+++ b/src/iana-types.json
@@ -605,6 +605,7 @@
     ]
   },
   "application/dicom": {
+    "extensions": ["dcm"],
     "sources": [
       "https://tools.ietf.org/rfc/rfc3240.txt",
       "https://www.iana.org/assignments/media-types/application/dicom"


DimitriPapadopoulos:
Looks like a dpulicate of #287.

pzsette:
@DimitriPapadopoulos How long does it take to approve the pull request?

dougwilson (changes requested):
Hello, and thank you for your contribution. Unfortunately the file you changed is a generated file and your changes would be overridden. Please see our contribution instructions at https://github.com/jshttp/mime-db#direct-inclusion

dougwilson:
Sorry, I thought I already commented on it, but apparently I never submitted the comment. The PR needs to be updated to change a non generated file.

DimitriPapadopoulos:
@dougwilson So #287 **is** the correct fix after all, with `"application/dicom"` split between `src/iana-types.json` for the part extracted as is from the IANA database, and the addition in `src/custom-types.json`?

Shouldn't the IANA database be fixed upstream?

wesleytodd:
Yes, I believe that if you resolve the issues in #287 that would be the correct update. That is unless since 2023 the upstream had been fixed to include the extension. I will leave it to you to decide what the next steps are for that one, but will close this one in the mean time.

PR #308, opened 2023-10-10
add automatic update GitHub Actions workflow

shogo82148:
It looks [apache-types.json](https://github.com/jshttp/mime-db/blob/master/src/apache-types.json) and [iana-types.json](https://github.com/jshttp/mime-db/blob/master/src/iana-types.json) are out dated.
We need to update them.
I have created a workflow to automate this operation.

An example of a pull request created by this workflow can be found here: https://github.com/shogo82148/mime-db/pull/1

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,26 @@
+name: update
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Node.js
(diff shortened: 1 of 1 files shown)


wesleytodd:
Hey! I am sorry I didn't see this sooner, but after doing some updates along with @jonchurch I think we ended up [doing what you recommended](https://github.com/jshttp/mime-db/blob/master/.github/workflows/generate.yml) without realizing this was open here. Sorry, it was not intentional but an oversight because the express project (which governs this package) has recently undergone some revitalization efforts and we just were not aware this was here. I am going to close this, but going forward the new folks would love more contributors like yourself so I hope this doesn't put you off from future contributions, I am sure we could use your input on other updates to this library or others in this org.

PR #309, opened 2023-10-20
Update db.json: Add missing JPEG extensions

shoonia:
Add missing JPEG extensions

- https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#jpeg_joint_photographic_experts_group_image

### DIFF
--- a/db.json
+++ b/db.json
@@ -7333,7 +7333,7 @@
   "image/jpeg": {
     "source": "iana",
     "compressible": false,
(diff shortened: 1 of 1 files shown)


dougwilson:
Hello, and thank you for you pull request. Unfortunately it cannot be accepted, as `db.json` is a generated file and manual edits will simply be overwritten.

If you would like to simply file an issue you are welcome to. If you want to contribute a change, please follow https://github.com/jshttp/mime-db#contributing

PR #312, opened 2024-01-02
Add shared-mime-types as upstream source

rimmington:
Adds Freedesktop.org's shared-mime-info repository as an upstream. shared-mime-info is used by common Linux desktop environments for MIME information, typically to identify which application to use to open a file. I've generally found sensible rationales for inclusion and changes in the commit messages of the repository.

There are quite a lot of `x-` types in this source, which reflects the upstream need for a file type for each format even if there is no IANA official or consensus type. Given that both Apache and Nginx also have a lot of `x-` types I think it's sensible to include them all.

### DIFF
--- a/db.json
+++ b/db.json
@@ -114,7 +114,12 @@
     "source": "iana",
     "extensions": ["ez"]
   },
+  "application/annodex": {
+    "source": "shared-mime-types",
+    "extensions": ["anx"]
+  },
   "application/appinstaller": {
+    "source": "shared-mime-types",
     "compressible": false,
     "extensions": ["appinstaller"]
   },
@@ -126,10 +131,12 @@
     "extensions": ["aw"]
   },
   "application/appx": {
+    "source": "shared-mime-types",
     "compressible": false,
     "extensions": ["appx"]
   },
   "application/appxbundle": {
+    "source": "shared-mime-types",
     "compressible": false,
     "extensions": ["appxbundle"]
   },
@@ -240,7 +247,8 @@
     "compressible": true
   },
   "application/cbor": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["cbor"]
   },
   "application/cbor-seq": {
     "source": "iana"
@@ -428,7 +436,8 @@
     "compressible": true
   },
   "application/dicom": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["dcm"]
   },
   "application/dicom+json": {
     "source": "iana",
@@ -481,7 +490,7 @@
   "application/ecmascript": {
     "source": "apache",
     "compressible": true,
-    "extensions": ["ecma"]
+    "extensions": ["ecma","es"]
   },
   "application/edi-consent": {
     "source": "iana"
@@ -612,7 +621,8 @@
     "compressible": true
   },
   "application/fits": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["fits","fit","fts"]
   },
   "application/flexfec": {
     "source": "iana"
@@ -655,6 +665,10 @@
     "compressible": true,
     "extensions": ["gml"]
   },
+  "application/gnunet-directory": {
+    "source": "shared-mime-types",
+    "extensions": ["gnd"]
+  },
   "application/gpx+xml": {
     "source": "apache",
     "compressible": true,
@@ -684,6 +698,10 @@
     "charset": "UTF-8",
     "compressible": true
   },
+  "application/hta": {
+    "source": "shared-mime-types",
+    "extensions": ["hta"]
+  },
   "application/http": {
     "source": "iana"
   },
@@ -704,6 +722,10 @@
   },
   "application/iges": {
     "source": "iana"
+  },
+  "application/illustrator": {
+    "source": "shared-mime-types",
+    "extensions": ["ai"]
   },
   "application/im-iscomposing+xml": {
     "source": "iana",
@@ -782,7 +804,8 @@
   },
   "application/jrd+json": {
     "source": "iana",
-    "compressible": true
+    "compressible": true,
+    "extensions": ["jrd"]
   },
   "application/jscalendar+json": {
     "source": "iana",
@@ -802,6 +825,7 @@
     "source": "iana"
   },
   "application/json5": {
+    "source": "shared-mime-types",
     "extensions": ["json5"]
   },
   "application/jsonml+json": {
@@ -909,7 +933,7 @@
   "application/mathml+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["mathml"]
+    "extensions": ["mathml","mml"]
   },
   "application/mathml-content+xml": {
     "source": "iana",
@@ -1075,10 +1099,12 @@
     "compressible": true
   },
   "application/msix": {
+    "source": "shared-mime-types",
     "compressible": false,
     "extensions": ["msix"]
   },
   "application/msixbundle": {
+    "source": "shared-mime-types",
     "compressible": false,
     "extensions": ["msixbundle"]
   },
@@ -1086,6 +1112,10 @@
     "source": "iana",
     "compressible": false,
     "extensions": ["doc","dot"]
+  },
+  "application/msword-template": {
+    "source": "shared-mime-types",
+    "extensions": ["dot"]
   },
   "application/mud+json": {
     "source": "iana",
@@ -1185,6 +1215,15 @@
   "application/oscore": {
     "source": "iana"
   },
+  "application/ovf": {
+    "source": "shared-mime-types",
+    "extensions": ["ova"]
+  },
+  "application/owl+xml": {
+    "source": "shared-mime-types",
+    "compressible": true,
+    "extensions": ["owx"]
+  },
   "application/oxps": {
     "source": "iana",
     "extensions": ["oxps"]
@@ -1226,15 +1265,15 @@
   "application/pgp-encrypted": {
     "source": "iana",
     "compressible": false,
-    "extensions": ["pgp"]
+    "extensions": ["pgp","gpg","asc"]
   },
   "application/pgp-keys": {
     "source": "iana",
-    "extensions": ["asc"]
+    "extensions": ["skr","pkr","asc","pgp","gpg","key"]
   },
   "application/pgp-signature": {
     "source": "iana",
-    "extensions": ["sig","asc"]
+    "extensions": ["sig","asc","pgp","gpg"]
   },
   "application/pics-rules": {
     "source": "apache",
@@ -1255,7 +1294,8 @@
     "extensions": ["p10"]
   },
   "application/pkcs12": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["p12","pfx"]
   },
   "application/pkcs7-mime": {
     "source": "iana",
@@ -1270,7 +1310,8 @@
     "extensions": ["p8"]
   },
   "application/pkcs8-encrypted": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["p8e"]
   },
   "application/pkix-attr-cert": {
     "source": "iana",
@@ -1365,7 +1406,12 @@
   "application/qsig": {
     "source": "iana"
   },
+  "application/ram": {
+    "source": "shared-mime-types",
+    "extensions": ["ram"]
+  },
   "application/raml+yaml": {
+    "source": "shared-mime-types",
     "compressible": true,
     "extensions": ["raml"]
   },
@@ -1379,7 +1425,7 @@
   "application/rdf+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["rdf","owl"]
+    "extensions": ["rdf","rdfs","owl"]
   },
   "application/reginfo+xml": {
     "source": "iana",
@@ -1508,6 +1554,11 @@
     "source": "iana",
     "compressible": true
   },
+  "application/schema+json": {
+    "source": "shared-mime-types",
+    "compressible": true,
+    "extensions": ["json"]
+  },
   "application/scim+json": {
     "source": "iana",
     "compressible": true
@@ -1633,7 +1684,7 @@
   "application/smil+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["smi","smil"]
+    "extensions": ["smi","smil","sml","kino"]
   },
   "application/smpte336m": {
     "source": "iana"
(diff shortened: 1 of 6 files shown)


db.json
```diff
@@ -428,7 +436,8 @@
     "compressible": true
   },
   "application/dicom": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["dcm"]
   },
```
jayvdb, line 441: This is correct - I think the quotes in https://www.iana.org/assignments/media-types/application/dicom are causing this extension to be omitted from mime-db at the moment.

db.json
```diff
@@ -612,7 +621,8 @@
     "compressible": true
   },
   "application/fits": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["fits","fit","fts"]
   },
```
jayvdb, line 626: correct per https://www.iana.org/assignments/media-types/application/fits

db.json
```diff
@@ -782,7 +804,8 @@
...
   "application/jrd+json": {
     "source": "iana",
-    "compressible": true
+    "compressible": true,
+    "extensions": ["jrd"]
   },
```
jayvdb, line 809: correct per https://www.iana.org/assignments/media-types/application/jrd+json

db.json
```diff
@@ -909,7 +933,7 @@
   "application/mathml+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["mathml"]
+    "extensions": ["mathml","mml"]
   },
```
jayvdb, line 937: correct per https://www.iana.org/assignments/media-types/application/mathml+xml

db.json
```diff
@@ -1379,7 +1425,7 @@
   "application/rdf+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["rdf","owl"]
+    "extensions": ["rdf","rdfs","owl"]
   },
```
jayvdb, line 1429: .rdfs is not mentioned at https://www.iana.org/assignments/media-types/application/rdf+xml

db.json
```diff
@@ -1633,7 +1684,7 @@
   "application/smil+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["smi","smil"]
+    "extensions": ["smi","smil","sml","kino"]
   },
```
jayvdb, line 1688: .sml correct per https://www.iana.org/assignments/media-types/application/smil , but .kino is not mentioned there

db.json
```diff
@@ -1647,7 +1698,7 @@
...
   "application/sparql-query": {
     "source": "iana",
-    "extensions": ["rq"]
+    "extensions": ["rq","qs"]
   },
```
jayvdb, line 1702: .qs not mentioned on https://www.iana.org/assignments/media-types/application/sparql-query

db.json
```diff
@@ -2111,7 +2163,8 @@
     "extensions": ["air"]
   },
   "application/vnd.adobe.flash.movie": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["swf","spl"]
   },
```
jayvdb, line 2168: https://www.iana.org/assignments/media-types/application/vnd.adobe.flash.movie mentions the commonly used .swf , but not .spl

db.json
```diff
@@ -2203,7 +2256,8 @@
     "extensions": ["azw"]
   },
   "application/vnd.amazon.mobi8-ebook": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["azw3","kfx"]
   },
```
jayvdb, line 2261: .azw3 mentioned on https://www.iana.org/assignments/media-types/application/vnd.amazon.mobi8-ebook , but not .kfx

db.json
```diff
@@ -2281,7 +2339,7 @@
...
   "application/vnd.apple.mpegurl": {
     "source": "iana",
-    "extensions": ["m3u8"]
+    "extensions": ["m3u8","m3u"]
   },
```
jayvdb, line 2343: correct per https://www.iana.org/assignments/media-types/application/vnd.apple.mpegurl

db.json
```diff
@@ -2424,7 +2483,8 @@
     "extensions": ["cdxml"]
   },
   "application/vnd.chess-pgn": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["pgn"]
   },
```
jayvdb, line 2488: correct per https://www.iana.org/assignments/media-types/application/vnd.chess-pgn

db.json
```diff
@@ -2472,7 +2532,8 @@
     "source": "iana"
   },
   "application/vnd.coffeescript": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["coffee"]
   },
```
jayvdb, line 2537: correct per https://www.iana.org/assignments/media-types/application/vnd.coffeescript

db.json
```diff
@@ -2506,10 +2567,12 @@
...
     "source": "iana",
-    "compressible": false
+    "compressible": false,
+    "extensions": ["cbz"]
   },
   "application/vnd.comicbook-rar": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["cbr"]
   },
```
jayvdb, line 2576: correct per https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar

db.json
jayvdb: It was added at https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/150637a3c0ec0b4f321bd49dc41650b8e71c559f - ``Closes: #8389`` refers to https://bugs.freedesktop.org/show_bug.cgi?id=8389 , which is thankfully still operational.  So in addition to commit messages, the background which led to each commit is likely also available.

fwiw, .rdfs is RDF Schema , and it is expressed as a RDF, and can be serialised as XML, so this addition is fine.

db.json
jayvdb: https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/aeec6ec3c98441db041e81274478c41dd9a38483 & https://bugs.freedesktop.org/show_bug.cgi?id=3141

This is about Linux only software discontinued in 2009 https://en.wikipedia.org/wiki/Kino_(software)

https://sourceforge.net/p/kino/code/HEAD/tree/trunk/smilutils/ and the manpage under that dont give any indication that .dino was a recommended extension.

However there are some hits on github for .kino being used for smil files https://github.com/search?q=path%3A*.kino&type=code

So this looks like a low value extension, but still valid.

db.json
jayvdb: Added in https://gitlab.freedesktop.org/xdg/shared-mime-info/-/merge_requests/176 ; I cant see any evidence for .qs being used for this, and [github results](https://github.com/search?q=path%3A*.qs&type=code) are drowned out with it being used for https://doc.qt.io/qt-5/qtscript-index.html and other scripting languages.

db.json
jayvdb: Added in https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/298543c7e449165b46a85d71209faf42de508217 - .sql is added with the media type of Futuresplash

https://www.webdesignmuseum.org/software/futuresplash-animator-in-1996 lists "FutureSplash Player SPL (Export)" under image formats, not animation formats, so I suspect that they are two different formats.

https://en.wikipedia.org/wiki/SWF lists .spl as a "Related file formats and extension"

https://github.com/ruffle-rs/ruffle/issues/18260 and https://github.com/ruffle-rs/ruffle/blob/295bd1062cfa281beeb4371f44b3cfb53c12171b/web/packages/core/src/swf-utils.ts#L24 says .spl is the same as .swf

First few pages of https://github.com/search?q=path%3A*.spl&type=code are not flash.

I am not 100% sure, but this seems to be ok.

db.json
```diff
@@ -3727,7 +3814,7 @@
...
   "application/vnd.lotus-1-2-3": {
     "source": "iana",
-    "extensions": ["123"]
+    "extensions": ["123","wk1","wk3","wk4","wks"]
```
jayvdb, line 3817: .wks not mentioned on https://www.iana.org/assignments/media-types/application/vnd.lotus-1-2-3 , but it is an early version of the file format.

db.json
```diff
@@ -3826,7 +3913,8 @@
     "extensions": ["igx"]
   },
   "application/vnd.microsoft.portable-executable": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["exe","dll","cpl","drv","scr","efi","ocx","sys","lib"]
```
jayvdb, line 3917: https://www.iana.org/assignments/media-types/application/vnd.microsoft.portable-executable only lists .exe and .dll 

Long commit message at https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/d4c2ed1b62b51fed1f9fa003de6cfdc6a82df729

I think .lib may be misguided - they are COFF files on Windows.

It looks like they are getting that long list of file extensions from the infobox on http://fileformats.archiveteam.org/wiki/Portable_Executable

db.json
```diff
@@ -3930,7 +4023,7 @@
   "application/vnd.ms-excel": {
     "source": "iana",
     "compressible": false,
-    "extensions": ["xls","xlm","xla","xlc","xlt","xlw"]
+    "extensions": ["xls","xlm","xla","xlc","xlt","xlw","xll","xld"]
   },
```
jayvdb, line 4027: https://www.iana.org/assignments/media-types/application/vnd.ms-excel is very strange

db.json
```diff
@@ -4275,13 +4404,17 @@
...
   "application/vnd.oasis.opendocument.formula-template": {
     "source": "iana",
-    "extensions": ["odft"]
+    "extensions": ["odft","otf"]
   },
```
jayvdb, line 4408: https://www.iana.org/assignments/media-types/application/vnd.oasis.opendocument.formula-template only lists .otf, not the pre-existing .otft

db.json
```diff
@@ -5274,7 +5419,7 @@
...
   "application/vnd.smaf": {
     "source": "iana",
-    "extensions": ["mmf"]
+    "extensions": ["mmf","smaf"]
```
jayvdb, line 5422: .smaf not on https://www.iana.org/assignments/media-types/application/vnd.smaf

Add in https://gitlab.freedesktop.org/xdg/shared-mime-info/-/commit/343b7b9a9304c322b3fc7dc60057de233f6e6662

None at https://github.com/search?q=path%3A*.smaf&type=code

I dont see a lot of evidence of .smaf being used

db.json
```diff
@@ -5311,7 +5460,12 @@
     "extensions": ["sfs"]
   },
   "application/vnd.sqlite3": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["sqlite3"]
+  },
```
jayvdb, line 5465: https://www.iana.org/assignments/media-types/application/vnd.sqlite3 suggests others also: .sqlite & .db

db.json
```diff
@@ -5686,7 +5848,7 @@
   },
   "application/vnd.wordperfect": {
     "source": "iana",
-    "extensions": ["wpd"]
+    "extensions": ["wpd","wp","wp4","wp5","wp6","wpp"]
   },
```
jayvdb, line 5852: https://www.iana.org/assignments/media-types/application/vnd.wordperfect only lists .wpd

db.json
```diff
@@ -6389,20 +7464,56 @@
...
+  },
   "application/x-www-form-urlencoded": {
     "source": "iana",
     "compressible": true
   },
   "application/x-x509-ca-cert": {
     "source": "iana",
-    "extensions": ["der","crt","pem"]
+    "extensions": ["der","crt","pem","cert"]
   },
```
jayvdb, line 7502: https://www.iana.org/assignments/media-types/application/x-x509-ca-cert lists 0 extensions

db.json
```diff
@@ -6486,7 +7622,7 @@
   "application/xhtml+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["xhtml","xht"]
+    "extensions": ["xhtml","xht","html","htm"]
   },
```
jayvdb, line 7626: https://www.iana.org/assignments/media-types/application/xhtml+xml lists only "xhtml","xht"

This is one where I feel that adding `"html","htm"` probably makes this mime-db less useful, as those file extensions are very commonly used, and very unlikely to be XML most of the time.

db.json
```diff
@@ -6495,20 +7631,21 @@
   "application/xliff+xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["xlf"]
+    "extensions": ["xlf","xliff"]
```
jayvdb, line 7634: https://www.iana.org/assignments/media-types/application/xliff+xml only lists xlf

But https://github.com/search?q=path%3A*.xliff&type=code shows a lot of positive hits

db.json
```diff
@@ -6495,20 +7631,21 @@
...
   },
   "application/xml": {
     "source": "iana",
     "compressible": true,
-    "extensions": ["xml","xsl","xsd","rng"]
+    "extensions": ["xml","xsl","xbl","xsd","rng"]
   },
```
jayvdb, line 7640: https://www.iana.org/assignments/media-types/application/xml only lists .xml 

.xbl is part of XUL , a UI language which failed standardisation, and doesnt have a lot of usage these days (sob).

db.json
```diff
@@ -6495,20 +7631,21 @@
...
   "application/xml-dtd": {
     "source": "iana",
     "compressible": true,
     "extensions": ["dtd"]
   },
   "application/xml-external-parsed-entity": {
-    "source": "iana"
+    "source": "iana",
+    "extensions": ["ent"]
   },
```
jayvdb, line 7649: https://www.iana.org/assignments/media-types/application/xml-external-parsed-entity says ".xml" or ".ent" :+1:

jayvdb:
A bit of a marathon - I've reviewed all the `application/` which currently have a source of "iana".  Initially I commented on a few to indicate they matched IANA records, but that was too many so I switched to only commenting when there was a discrepancy.  I have analysed most of the discrepancies - while a few I feel are adding too many extensions, in most cases these are harmless, as there are no other media/mime types using that extension.

`.lib` and `.htm` and `.html` seem like the most problematic

wesleytodd:
Wow thanks @jayvdb for the review and @rimmington for the PR. I think we need to consider how we want to approach a new source in terms of semver-ness here. AFAIK this would be the first such additional source and therefore likely needs a bit more consideration than *just* the data changes (not to say those are not also critical, and thanks again for reviewing those @jayvdb). I am not sure I have a really strong opinion, but to be sure we get a few more eyes on this I will also ping in out @jshttp/express-tc.

bjohansebas:
I see this PR for a future version 2 of mime-db, which I would think would be ideal, before adding it to version 1.

PR #323, opened 2024-05-05
Fix: Update audio/webm extension

xNevrroXx:
The audio/webm mime type extension was incorrect.

### DIFF
--- a/src/apache-types.json
+++ b/src/apache-types.json
@@ -2700,7 +2700,7 @@
   "audio/vorbis": {},
   "audio/vorbis-config": {},
   "audio/webm": {
-    "extensions": ["weba"]
+    "extensions": ["webm"]
   },
   "audio/x-aac": {
     "extensions": ["aac"]


UlisesGascon:
Hi @xNevrroXx! Thanks for proposing this change! :+1:  

I was checking some [online resources](https://fileinfo.com/extension/weba) and seems like `.weba` is only audio (`audio/webm`) while `.webm` is video (`video/webm`). 

Can you provide additional context for this change? :)

xNevrroXx:
Hi, @UlisesGascon
Maybe you are right. Some sites indicate that `.weba` is an extension of the `audio/webm` mime type. But [others](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Containers#browser_compatibility) indicate both `audio/webm` and `video/webm` also have the extension `.webm`.
But, I checked out the ability of browsers to play `audio-name.weba` files - they can play them exactly the same as `audio-name.webm`. But if we will talk about Windows OS 11 (imagine that we are trying to download some files to the client side) - this one cannot play `audio-name.weba`, but can play `audio-name.webm` files with native music player. 

So, it's the hard question)
Thanks for your attention

dougwilson:
Unless upsteam Apache actually changed the mapping, your change will be overwritten because you changed a generated file. You should contact Apache project for changes to Apache mappings.

xNevrroXx:
@UlisesGascon 
It seems that [Wikipedia](https://en.wikipedia.org/wiki/Audio_file_format) doesn't know about `.weba` audio extension either.

PR #324, opened 2024-05-29
fix: make `application/vnd.unity` compressible

styfle:
The Unity documentation mentions both gzip and brotli are recommended when serving `.unityweb` files.

https://docs.unity3d.com/Manual/webgl-deploying.html

### DIFF
--- a/db.json
+++ b/db.json
@@ -5530,6 +5530,7 @@
   },
   "application/vnd.unity": {
     "source": "iana",
+    "compressible": true,
     "extensions": ["unityweb"]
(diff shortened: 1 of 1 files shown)


styfle:
@dougwilson @UlisesGascon Is this the correct file to edit when adding `compression`? Let me know if I need to change anything, thanks!

creationix (commented):
We probably don't want to mark this as compressible.  The file format can be pre-compressed already and it doesn't change the file extension in all cases.

PR #325, opened 2024-06-01
Added `.gpkg` extension for `application/geopackage+sqlite3` from IANA

ThorodanBrom:
Hello, this PR adds the extension `.gpkg` for GeoPackage files MIME type `application/geopackage+sqlite3`. The extension is defined by IANA here - https://www.iana.org/assignments/media-types/application/geopackage+sqlite3

The commit is more or less similar to what has been done in 48e411522aed9925570b6c6bbf0bbc331242c038.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Add extension `gpkg.` to `application/geopackage+sqlite3`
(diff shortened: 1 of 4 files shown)


wesleytodd (changes requested):
We cannot land this PR as is, please follow the guides for which files to update manually. The ones you updated here should not be manually edited.

PR #326, opened 2024-06-17
new apache type "parquet"

johannesrue:
Hey,
i have added a new type `application/vnd.apache.parquet`. 

Cheers
Johannes


wesleytodd (changes requested):
These files are generated. Please see the section of the readme for including types which are not from the upstream

PR #327, opened 2024-07-11
create a GHA to run update and open a PR if diff

jonchurch:
Add a Github Action to run `npm run update` on the first of each month and create a PR if there are diffs here.

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -1,0 +1,58 @@
+name: Update and Create PR
+
+on:
+  schedule:
+    - cron: '0 0 1 * *' # Runs on the first day of each month at midnight
+  workflow_dispatch: # Allows the workflow to be triggered manually
+
+jobs:
+  update-and-pr:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate.yml
```diff
@@ -0,0 +1,58 @@
...
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          if [ -n "$(git status --porcelain)" ]; then
+            BRANCH_NAME="update-branch"
+            if git show-ref --verify --quiet "refs/heads/$BRANCH_NAME"; then
+              TIMESTAMP=$(date +%s)
+              BRANCH_NAME="${BRANCH_NAME}-${TIMESTAMP}"
+            fi
+            git checkout -b $BRANCH_NAME
+            git add .
```
jonchurch, line 39: not the most elegant thing to do here, since this should only ever update `db.json` and `src/`

Thoughts?

jonchurch: Made this change

wesleytodd (approved):

PR #330, opened 2024-07-18
docs: notice about semver and data sources

wesleytodd:
As mentioned in #328. This documents the operational model which this package has always operated with. Yes we recognize this means data provided by this package can break applications, but as the notice says, if this is a concern to you please lock your version.


bjohansebas (approved):

UlisesGascon (approved):
LGTM! Not sure It took me that long to review it :face_in_clouds:

PR #344, opened 2024-11-22
Add support for the x-compressed and x-zip-compressed mime types

pedro-php:
Summary:
This PR adds support for the x-compressed and x-zip-compressed mime types based of Nginx servers

application/x-compressed
application/x-zip-compressed

### DIFF
--- a/src/nginx-types.json
+++ b/src/nginx-types.json
@@ -83,6 +83,9 @@
   "application/x-cocoa": {
     "extensions": ["cco"]
   },
+  "application/x-compressed:":{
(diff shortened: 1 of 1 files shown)


wesleytodd:
These files are generated: https://github.com/jshttp/mime-db/blob/master/scripts/fetch-nginx.js

It does look like we missed that our automation was failing: https://github.com/jshttp/mime-db/actions/runs/11622470327/job/32367979107

Maybe you could look at that and see if re-running the nginx fetch script will pull these in?

PR #346, opened 2024-11-25
Update custom-types.json

pedro-php:
Summary:
This PR adds support for compressed types that, despite being deprecated by iana, are still being used by windows.

Added MIME Types:

Mime: application/x-zip-compressed 
Extension: zip
Notes: Windows use this mime-type for Zip files instead of application/zip.
Sources: (https://bugs.python.org/issue41738)
               (https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types/Common_types)

application/x-compressed
Extension: rar
Notes: Windows 11 use this mime-type for RAR files instead of application/x-rar-compressed or application/vnd.rar

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -317,6 +317,18 @@
   "application/x-chrome-extension": {
     "extensions": ["crx"]
   },
+  "x-compressed": {
+    "extensions": ["rar"],
+    "notes": "Windows 11 use this mime-type for RAR files instead of application/x-rar-compressed or application/vnd.rar"
+  },
+  "x-zip-compressed": {
+    "extensions": ["zip"],
+    "sources": [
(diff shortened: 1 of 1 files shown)


wesleytodd (approved):
Hm, I am not sure how I feel about these. I don't really see anything wrong with adding them if they are used in the wild, but if they are deprecated then I am also not sure why adding them now add's value.

I will add an approval and see if we can get some more feedback on these. Feel free to post more details about the impact of not including these so we can decide if we wan to merge this.

UlisesGascon (approved):

bjohansebas (approved):

wesleytodd:
Ok, got a few more approvals. Merging.

PR #347, opened 2024-11-27
fix workflow

bjohansebas:
This fixes the automation. The issue was that `got` throws an exception when it returns a status code outside of the 2xx and 3xx range.

Fun fact: while trying to fix this with `undici`, the PR size was larger, but with this change, the PR is smaller (The one that generates the automation).

### DIFF
--- a/scripts/fetch-iana.js
+++ b/scripts/fetch-iana.js
@@ -94,24 +94,28 @@ async function addTemplateData (data, options) {
     return
   }
 
-  let res = await got('https://www.iana.org/assignments/media-types/' + data.template)
+  let res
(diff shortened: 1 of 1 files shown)


bjohansebas:
Btw, I can migrate to Undici, which still supports CommonJS and is faster than Got.

UlisesGascon (approved):

UlisesGascon:
I love more the idea of using Undici https://github.com/jshttp/mime-db/pull/352, but anyway I am happy to land this change first in case that we don't want to change library. Both options are fine for me. Thanks @bjohansebas for doing this!

bjohansebas:
Close in favor of #355

PR #352, opened 2024-12-13
replace got with undici

bjohansebas:
Unlike what #347 does, got is replaced with undici, avoiding the use of try-catch, with the same result. It might improve the workflow speed, but that's the least important thing (in my opinion).

### DIFF
--- a/scripts/fetch-iana.js
+++ b/scripts/fetch-iana.js
@@ -9,7 +9,7 @@
  * Convert the IANA definitions from CSV to local.
  */
 
-var { request } = require('./utils')
+var { request } = require('./lib/request')
 var parser = require('csv-parse')
 var toArray = require('stream-to-array')
 var typer = require('media-typer')
--- a/scripts/fetch-apache.js
+++ b/scripts/fetch-apache.js
@@ -12,7 +12,7 @@
  */
 
 var writedb = require('./lib/write-db')
-var { request } = require('./utils')
+var { request } = require('./lib/request')
 
 /**
  * Mime types and associated extensions are stored in the form:
--- a/scripts/fetch-nginx.js
+++ b/scripts/fetch-nginx.js
@@ -11,7 +11,7 @@
  * Convert these text files to JSON for browser usage.
  */
 
-var { request } = require('./utils')
+var { request } = require('./lib/request')
 var writedb = require('./lib/write-db')
 
 /**
--- a/scripts/lib/request.js
+++ b/scripts/lib/request.js
@@ -1,5 +1,6 @@
 const { request } = require('undici')
 
+// Undici does not send the user-agent header, which causes the request to fail for the IANA script.
 module.exports.request = async function (url, options) {
   return request(url, {
     ...options,


scripts/fetch-iana.js
```diff
@@ -213,13 +215,13 @@ function extractTemplateExtensions (body) {
...
 async function get (type, options) {
-  const res = await got('https://www.iana.org/assignments/media-types/' + encodeURIComponent(type) + '.csv')
+  const res = await request('https://www.iana.org/assignments/media-types/' + encodeURIComponent(type) + '.csv')
 
   if (res.statusCode !== 200) {
     throw new Error('got status code ' + res.statusCode + ' from ' + type)
   }
 
-  const mimes = await toArray(parser(res.body))
+  const mimes = await toArray(parser(await res.body.text()))
```
UlisesGascon, line 224:
```suggestion
  const responseText = await res.body.text();
  const mimes = await toArray(parser(responseText));
```

I believe this approach improves readability and makes the code easier to maintain.

UlisesGascon (approved):
LGTM! I like the idea of using undici in this scenario :+1:

scripts/fetch-iana.js
bjohansebas: Done

UlisesGascon:
Not sure why the coverage is failing, but does not seems related (IMO) https://github.com/jshttp/mime-db/actions/runs/12635017301/job/35239389864?pr=352

scripts/fetch-iana.js
```diff
@@ -123,11 +125,11 @@ async function addTemplateData (data, options) {
     throw new Error('got status code ' + res.statusCode + ' from template ' + data.template)
   }
 
-  var body = getTemplateBody(res.body)
+  var body = getTemplateBody(await res.body.text())
   var mime = extractTemplateMime(body)
 
   // add the template as a source
-  addSource(data, res.url)
```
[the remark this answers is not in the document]
bjohansebas: Undici does not provide the `url` property, so I had to create a variable and update it manually to replicate the behavior.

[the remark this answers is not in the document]
bjohansebas: We're also no longer receiving redirects since I changed the URL that was causing the redirect in:

https://github.com/jshttp/mime-db/blob/e7548fb503f2f1015026c09012b47e2abe7a82a6/scripts/fetch-iana.js#L113-115

scripts/utils.js
```diff
@@ -0,0 +1,8 @@
+const { request } = require('undici')
```
wesleytodd, line 1: I am not sure I understand the need for this file? Do we need this user agent at all? If so, we should comment about it because it seems like this entire file exists just to set that, otherwise we could just import `unidici` directly into each of the scripts.

EDIT: I should add I consider any file/directory with `util` as the name a "smell". Meaning it was not well thought out. In this case, the file should be called `request` if you really do need it.

wesleytodd (changes requested):

scripts/utils.js
bjohansebas: This file is necessary because, for some reason, the user-agent header is not being sent, and it is only for configuration. It is required to make requests to IANA, otherwise, the request would fail without that header.

wesleytodd: Ok, can you document that reason in a comment in this file? With that change I will move it to an approval.

bjohansebas: done

wesleytodd (approved):

scripts/fetch-iana.js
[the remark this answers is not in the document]
wesleytodd: @jonchurch do you consider this resolved? If so I would like to merge this.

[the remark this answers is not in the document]
wesleytodd: @jonchurch friendly ping on this.

jonchurch (approved):

PR #354, opened 2024-12-23
chore: upgrade scorecard workflow pinned action versions

carpasse:
Upgrade scorecard workflow pinned action versions to latest versions.

Issue https://github.com/expressjs/security-wg/issues/2

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -32,12 +32,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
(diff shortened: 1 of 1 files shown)


UlisesGascon (approved):

blakeembrey:
Can we move these into another repo instead? https://github.com/expressjs/security-wg/issues/31

Why are we upgrading? What’s changed? This doesn’t fix that scorecard workflows turn themselves off when the repo isn’t popular right?

wesleytodd:
I agree with @blakeembrey. I dont see a reason to close this until we get together a plan that doesn't involve so much distribute work. Centralizing it seems like a great way, and I think this is part of the next round of already planned work right?

UlisesGascon:
So far seems like centralization is not simple in this case https://github.com/expressjs/security-wg/issues/31#issuecomment-2813048690, so I will like to unblock this PR :smile:

Issue #345, opened 2024-11-23
Automation is broken


bjohansebas:
The automation is broken because it cannot correctly retrieve https://www.iana.org/assignments/media-types/image.csv, even though the endpoint is correct.

If that resource is removed for generation, the following PR https://github.com/bjohansebas/mime-db/pull/1 would be obtained, although ideally, it shouldn’t be removed.

I’ve added console.log to debug the issue, please check the results of the following workflows:

https://github.com/bjohansebas/mime-db/actions/runs/11982093952/job/33409524328 (work)
https://github.com/bjohansebas/mime-db/actions/runs/11982462369/job/33410514027 (not work)

bjohansebas:
@wes I think this is going to be a problem with got, and we can't update it because Got is now ESM-only. Maybe I can make this work again with Undici.

bjohansebas:
@wesleytodd I think this is going to be a problem with got, and we can't update it because Got is now ESM-only. Maybe I can make this work again with Undici.

jonchurch:
When the update script was refactored to use `got` in 92715b33 it didn't factor in that `got` throws on non success http status codes

So the built in 404 handling in the `addTemplateData` function will never run and we exit early on the unhandled promise rejection.

A workaround is to set `throwHttpErrors: false` when fetching

```js
let res = await got(url, { throwHttpErrors: false })
```

But the script needs a look over, since the assumptions about the http client's behavior are not valid anymore.

PR #355, opened 2025-01-04
dont throw on 404s when fetching templates from iana

jonchurch:
workaround for #345

When the update script was refactored to use `got` in 92715b33 it didn't factor in that `got` throws on non success http status codes

So the built in 404 handling in the `addTemplateData` function will never run and we exit early on the unhandled promise rejection.

A workaround is to set [`throwHttpErrors: false`](https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#throwhttperrors) when fetching

```js
let res = await got(url, { throwHttpErrors: false })
```

But the script needs a look over, since the assumptions about the http client's behavior are not valid anymore.

## Why Now?

there were deprecated unofficial (extension) mime types which are 404ing now, `image/x-emf, image/x-wmf` 
I don't see when they went standard as `image/emf, image/wmf`, or when the extension versions started 404ing

### DIFF
--- a/scripts/fetch-iana.js
+++ b/scripts/fetch-iana.js
@@ -94,7 +94,7 @@ async function addTemplateData (data, options) {
     return
   }
 
-  let res = await got('https://www.iana.org/assignments/media-types/' + data.template)
+  let res = await got(`https://www.iana.org/assignments/media-types/${data.template}`, { throwHttpErrors: false })
   var ref = data.type + '/' + data.name
   var rfc = getRfcReferences(data.reference)[0]
 


jonchurch:
We don't have tests for the update script Im realizing

scripts/fetch-iana.js
```diff
@@ -94,7 +94,7 @@ async function addTemplateData (data, options) {
     return
   }
 
-  let res = await got('https://www.iana.org/assignments/media-types/' + data.template)
+  let res = await got(`https://www.iana.org/assignments/media-types/${data.template}`, { throwHttpErrors: false })
```
UlisesGascon, line 97: Actually we don't support new ES6+ syntax based on the CI: https://github.com/jshttp/mime-db/blob/master/.github/workflows/ci.yml#L13

jonchurch: This script has es6 via async/await and let already, so the template string isn't a breaking introduction.

Interesting wrinkle here though, in that these scripts are part of the build tooling and we don't document the node version requirement for development/building. v11 of got has engines set to `10.19.0` https://github.com/sindresorhus/got/blob/1f6ac4597b797e6fe760a7dc11a3db8bf298aa94/package.json#L10C1-L11C1

I set up the automation PR that runs the build step on cron to use LTS node, so es6 is fine here.

https://github.com/jshttp/mime-db/blob/49f8df0e170c7b40785e8aa4f464793b7fcfa41b/.github/workflows/generate.yml#L23C1-L24C1

wesleytodd:
Is this superseded by #352? or still relevant?

jonchurch:
They both will solve the 404 problem, and if we decide to switch to undici then this pr is not needed

wesleytodd:
I prefer switching. Ideally at some point most of `undici`'s apis will be available directly from node, so then maybe we will even be able to drop the dep entirely. We could today with `fetch` maybe but I think that is not ideal for a few reasons (but we could discuss it in the future if we thought it was worth considering).

PR #356, opened 2025-01-09
ci: update ci

bjohansebas:
The CI from compression is copied, which doesn't fail coverage. I hope this solves it.


bjohansebas:
Well, now I just have to make it work again with Node 0.6, the coverage works fine.

wesleytodd (approved):

PR #358, opened 2025-01-21
Add custom type text/x-php 

cjiso1117:
Add support for `text/x-php` since `file` command use `text/x-php` in debian system

ref: https://cweiske.de/tagebuch/php-mimetype.htm#linux

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -905,6 +905,10 @@
       "http://orgmode.org"
     ]
   },
+  "text/x-php": {
+    "compressible": true,
+    "extensions": ["php"]
+  },
   "text/x-processing": {
     "compressible": true,
     "extensions": ["pde"],


cjiso1117:
is PR review stuck due to broken automation?

wesleytodd:
No I think it was stuck due to no one reviewing it. That said, if you [look here](https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types/Common_types) you see `application/x-httpd-php` which is already in our data. That article is from 2011 and I am not sure it justifies adding this type.

cjiso1117:
But `file` command return `text/x-php` for most os in 2025, and some mimetype guesser depend on `file` command and system's mimetype database.

bjohansebas (approved):

PR #361, opened 2025-03-05
Add MIME Types for Apple .ipa and .plist extensions

danilo-santana-e-silva:
From: Distribute proprietary in-house apps to Apple devices
Ref: https://support.apple.com/en-gb/guide/deployment/depce7cefc4d/1/web#dep30d16db76

"You may need to configure your web server so the manifest file and app file are transmitted correctly. For the server, add the MIME types to the web service’s MIME types settings:

application/octet-stream ipa
text/xml plist"

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -128,9 +128,10 @@
   },
   "application/octet-stream": {
     "compressible": false,
-    "extensions": ["buffer"],
+    "extensions": ["buffer", "ipa"],
     "sources": [
-      "https://github.com/broofa/node-mime/blob/v1.2.11/types/mime.types#L154"
+      "https://github.com/broofa/node-mime/blob/v1.2.11/types/mime.types#L154",
+      "https://support.apple.com/en-gb/guide/deployment/depce7cefc4d/1/web#dep30d16db76"
     ]
   },
   "application/ogg": {
@@ -939,7 +940,8 @@
     ]
   },
   "text/xml": {
-    "compressible": true
+    "compressible": true,
+    "extensions": ["plist"]
   },
   "text/yaml": {
     "compressible": true,


bjohansebas (approved):

src/custom-types.json
```diff
@@ -939,7 +940,8 @@
     ]
   },
   "text/xml": {
-    "compressible": true
+    "compressible": true,
+    "extensions": ["plist"]
```
wesleytodd, line 944: Hm, actually I am not sure what the impact is of adding this extension here. I *think* this is incorrect, but would like more eyes here. Maybe I am missing something.

bjohansebas: I think it would be safe to add it, this is the diff it makes:
image

wesleytodd: I don't see any precedent for this though. There are bunches of examples of files with custom extention's which are actually another format (xml, json, etc) that are not doing this. Why would we make a special case for this one?

wesleytodd: Yeah, and based on the docs from apple, it looks like they should have used a [suffix](https://en.wikipedia.org/wiki/Media_type#Suffix) and a specific mime type, not call this a `text/xml`. That would have made it more clear to me what is going on, and you can see plenty of examples of this in our current dataset. For example:

```
  "application/atom+xml": {
    "source": "iana",
    "compressible": true,
    "extensions": ["atom"]
  },
```

It is clear to me that they have documented it this way, and that as long as we prefer the `xml` extension first that it should be alright to add for the use cases *we have in the express project*. That said, I am concerned enough with this that I would like to ping in @dougwilson and get his opinion. Maybe I am being overly concerned, but since I cannot find another example like this I want to be extra careful.

wesleytodd: Ah! I was reviewing another PR and I think this might be a similar case to this one:

```
  "application/octet-stream": {
    "source": "iana",
    "compressible": false,
    "extensions": ["bin","dms","lrf","mar","so","dist","distz","pkg","bpk","dump","elc","deploy","exe","dll","deb","dmg","iso","img","msi","msp","msm","buffer"]
  },
```

As you can see here, lots of different extentions listed like you are adding to this one. I had been looking specifically at `xml` examples (and did a quick search for json when I couldn't find any xml ones) but this seems to me to be the same use case, and so sets the precident I was looking to find.

orionmind: I'm actually blocked on this limitation while using GH Pages to serve manifest.plist files for OTA Ad-Hoc Distribution of iOS apps.  The iOS user needs to hit a hyperlink using Apple's itms-services URL scheme (eg. itms-services://?action=download-manifest&url=https://myapp.com/manifest.plist).

Instead of an application/xml or text/xml content-type in the response, I'm seeing 'application/octet-stream' returned when this URL is visited, which makes the OTA Ad-Hoc installation fail.

bjohansebas:
Hey @danilo-santana-e-silva, could you please resolve the conflicts?

PR #362, opened 2025-03-05
Update mime type for DCM format

DimitriPapadopoulos:
Supersedes #287.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -515,7 +515,8 @@
   },
   "application/dicom": {
     "source": [
-        "https://en.wikipedia.org/wiki/DICOM"
+        "https://www.rfc-editor.org/rfc/rfc3240",
+        "https://www.dicomstandard.org/standards/view/media-formats-and-physical-media-for-media-interchange"
       ],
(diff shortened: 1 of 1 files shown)


bjohansebas (changes requested):
This should come from IANA.

https://www.iana.org/assignments/media-types/application/dicom

wesleytodd:
https://github.com/jshttp/mime-db/blob/master/db.json#L453-L463

I believe the reason this is being done is to extend the data with the extension which is not coming from iana.

wesleytodd (approved):

bjohansebas (approved):

DimitriPapadopoulos:
Indeed `application/dicom` is already part of [`src/iana-types.json`](https://github.com/jshttp/mime-db/blob/master/src/iana-types.json):
https://github.com/jshttp/mime-db/blob/22f89eda9fbc1ce3c561605fcaa9638a5a7cad47/src/iana-types.json#L645-L649

This is about associating the `dcm` file name extension to `application/dicom`.

PR #365, opened 2025-03-17
fix: compressible was set in the wrong file in #163

wesleytodd:
This fixes the fact that we missed that `compressible` was being changed in the generated file *not* the correct place.

### DIFF
--- a/src/custom-types.json
+++ b/src/custom-types.json
@@ -127,7 +127,7 @@
     ]
   },
   "application/octet-stream": {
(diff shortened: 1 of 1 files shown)


wesleytodd:
My editor did some autoformatting I guess. will push a fix.

wesleytodd:
Requesting a fast track once I get a single approval, since this is just fixing an error made in an already landed pr.

UlisesGascon (approved):

PR #371, opened 2025-05-13
Add log files support

ChaosAttractor:


bjohansebas:
Can you give us more information about this MIME?

ChaosAttractor:
I apologize for the wait.

`.log` files are associated with `text/x-log` if the file implements `Express.Multer.File` (express, nest.js file handler)

image

and `text/plain` if

```sh
file --mime-type -b test.log
# output will be text/plain and inode/x-empty if file is empty
```

or

```js
import path from 'path';
import mime from 'mime-types';

const ext = path.extname('./test.log');
const mimetype = mime.lookup(ext);

console.log(ext, mimetype);
```

Issue #397, opened 2025-08-04
mark application/vnd.pmtiles as not compressible


bdon:
I'm the maintainer of a [JavaScript implementation](https://github.com/protomaps/PMTiles/tree/main/js) of the [pmtiles specification](https://github.com/protomaps/PMTiles/blob/main/spec/v3/spec.md).

Recently storing`.pmtiles` on GitHub Pages and accessing them via HTTP Range Requests stopped working only in Firefox, because Firefox was accepting `gzip` content and GitHub Pages was serving a compressed response, leading to unreadable data. Full description here: https://github.com/protomaps/PMTiles/issues/584

According to GH support, Pages decides to compress responses based on MIME types in this DB. Renaming the `.pmtiles` file to `.pmtiles.gz` is a workaround, but does not fix existing Pages sites. 

I would like to add a custom entry to `src/custom-types.json` defining `.pmtiles` as not compressible.

PR #398, opened 2025-08-04
add application/vnd.pmtiles to custom-types.json [#397]

bdon:
See #397


db.json
bjohansebas: Please revert this change, you don’t need to do it, since our automation will take care of that

Issue #463, opened 2026-05-26
Add `.accdb` extension for application/x-msaccess (MS Access 2007+)


firefoxNX:
Is it possible to associate the `accdb` extension with a Microsoft Access media type?

  `mime-db` currently maps only the legacy `.mdb` extension:
  
  ```json
  "application/x-msaccess": {
    "source": "apache",
    "extensions": ["mdb"]
  } 
```

(If you'd prefer a distinct custom entry instead of extending the apache-sourced one, an alternative is a new application/x-msaccess / application/vnd.ms-access record in src/custom-types.json with "extensions": ["accdb"] and the source below.)

  Microsoft's official documentation describes .accdb as the Access database file format
  and uses the x-msaccess content type for Access database downloads:

  - Microsoft Learn - Access file formats / .accdb:
[  https://support.microsoft.com/en-us/office/which-access-file-format-should-i-use-cf6d6358-6 percentplaceholder](https://support.microsoft.com/en-us/access/which-access-file-format-should-i-use)

firefoxNX:
see https://github.com/jshttp/mime-db/pull/464

krzysdz:
The article that you have referenced does not mention the MIME type anywhere. Looking on the internet I found `application/x-msaccess` referring only to `.mdb` files. `.accdb` does not have MIME type listed on [Wikidata](https://www.wikidata.org/wiki/Q63391719) or the [Library of Congress](https://www.loc.gov/preservation/digital/formats/fdd/fdd000463.shtml).

Looking for `.accdb` MIME type I found only `application/msaccess` and all of the references seem to originate from Microsoft:

- .NET uses [`application/msaccess` for `.accdb`, `.accde`, `.accdt`](https://github.com/dotnet/runtime/blob/6963a5964693f9c439d100b4bba9ae01af697dc1/src/libraries/System.Net.Mail/src/System/Net/Mime/MediaTypeMap.cs#L29-L31), but reverse mapping from MIME to extension uses [`.mdb` for `application/msaccess`](https://github.com/dotnet/runtime/blob/6963a5964693f9c439d100b4bba9ae01af697dc1/src/libraries/System.Net.Mail/src/System/Net/Mime/MediaTypeMap.cs#L446).
  The mapping probably has [sources in IIS](https://github.com/microsoft/referencesource/blob/ec9fa9ae770d522a5b5f0607898044b7478574a3/System.Web/MimeMapping.cs#L98-L106).
[...]

PR #464, opened 2026-05-26
feat: add accdb extension to application/x-msaccess

firefoxNX:
## Summary

Microsoft Access 2007+ databases use the `.accdb` extension (the default Access file format since Access 2007), but `mime-db` only maps the legacy `.mdb` extension to `application/x-msaccess`.

As a result, `mime.lookup('file.accdb')` returns `false` and callers fall back to `application/octet-stream`, which breaks extension/MIME-based handling (content classification, conversion, downloads) for Access files.

This adds `accdb` to the existing `application/x-msaccess` entry, keeping it consistent with the current `.mdb` mapping.

## Change

`src/custom-types.json`:

```json
"application/x-msaccess": {
  "extensions": ["accdb"],
  "notes": "Microsoft Access 2007+ database file format (.accdb), successor to .mdb",
  "sources": [
    "https://support.microsoft.com/en-us/office/which-access-file-format-should-i-use-012d9ab3-d14c-479e-b617-be66f9070b41"
  ]
}
```

After `npm run build`, `db.json` resolves to:

```json
"application/x-msaccess": { "source": "apache", "extensions": ["mdb", "accdb"] }
```

## Source

[Microsoft — Which Access file format should I use?](https://support.microsoft.com/en-us/office/which-access-file-format-should-i-use-012d9ab3-d14c-479e-b617-be66f9070b41) — documents `.accdb` as the Microsoft Access database file format (introduced in Access 2007, successor to `.mdb`).

Note: `application/x-msaccess` is the convention already used in mime-db for `.mdb`; this PR extends it rather than introducing `application/vnd.ms-access`. Happy to switch to a distinct entry if preferred.

## Checklist

- [x] Added to `src/custom-types.json`
- [x] Ran `npm run build` to regenerate `db.json`
- [x] `npm run lint` passes
- [x] `npm test` passes


krzysdz (changes requested):
This is not the right MIME type for `.accdb`. The source does not mention any MIME type. See https://github.com/jshttp/mime-db/issues/463#issuecomment-4587111226 for more details.

firefoxNX:
Thanks @krzysdz, you're right — `application/x-msaccess` is the `.mdb` type and isn't linked to `.accdb` by any source, and the support article I originally cited doesn't state a MIME type.

I've updated the PR to use **`application/msaccess`** for `.accdb` (plus `.accde` and `.accdt`, which Microsoft's mappings cover identically), reverting `application/x-msaccess` back to `.mdb` only.

Source: the archived Microsoft MIME-types reference (Microsoft-authored, learn.microsoft.com), which maps all three to `application/msaccess` — consistent with the IIS / .NET / ASP.NET Core mappings you linked:
https://learn.microsoft.com/en-us/archive/blogs/jaimer/mime-types

`npm run build` re-run, lint + tests pass. Let me know if you'd prefer `.accdb` only rather than all three extensions.

firefoxNX:
Sorry. The archived blog carries a "no longer maintained" disclaimer — swapped the citation to the live, maintained Microsoft source-code mappings you referenced, which definitively map `.accdb`/`.accde`/`.accdt` to `application/msaccess`:

- .NET runtime `MediaTypeMap.cs`
- ASP.NET Core `FileExtensionContentTypeProvider.cs`

(both derived from the IIS default static MIME map). Updated in the latest commit.

